PR #1, opened 2015-08-04
Add usb device info script

kb100:
I think that a lot of people will find this script useful. Feel free to make pre-merge suggestions, but short of unforseen bugs I believe it is ready as is.


usb/usb.py
```diff
@@ -0,0 +1,245 @@
...
+    #    return True
+
+    # E.g. how to ignore a fixed set of paths
+    #if path in [ "/dev/path1", "/dev/path2", "/dev/path3" ]:
+    #    return True
+    return False
+
+###############################################################################
+# END CONFIG
+# DO NOT EDIT ANYTHING AFTER THIS POINT UNLESS YOU ARE A DEVELOPER
```
Zopieux: UNLESS YOU *KNOW WHAT YOU ARE DOING* :wink:

usb/usb.py
```diff
@@ -0,0 +1,245 @@
...
+    lines = lines.split("\n")
+    if len(lines) != 3:
+        return ""
+    else:
+        return lines[1].strip()
+
+def getLockedCryptOutput(path):
+    form = "<span color='{}'>[{} {}]</span>"
+    kname = getKernelName(path)
+    output = form.format(LOCKED_COLOR, FA_LOCK, kname)
```
Zopieux: `pangoEscape` is not used here. Is it on purpose?

usb/usb.py
```diff
@@ -0,0 +1,245 @@
...
+            spaceAvail = getSpaceAvailable(path)
+        mountPoint = "<i>{}</i>:".format(pangoEscape(mountPoint))
+    else:
+        color = UNLOCKED_NOT_MOUNTED_COLOR
+        spaceAvail = ""
+    kernelName = getKernelName(path)
+    parentKernelName = getParentKernelName(path)
+    
+    block = "<span color='{}'>[{} {}:{}]</span>"
+    block = block.format(color, FA_UNLOCK, parentKernelName, kernelName)
```
Zopieux: Same remark here on `pangoEscape`.

usb/usb.py
```diff
@@ -0,0 +1,245 @@
...
+    if label:
+        label = '"{}"'.format(label)
+
+    items = [block, label, mountPoint, spaceAvail]
+    return " ".join(filter(None, items))
+
+def getDiskWithNoPartitionsOutput(path):
+    form = "<span color='{}'>[{}] {}</span>"
+    kernelName = getKernelName(path)
+    return form.format(PARTITIONLESS_COLOR, kernelName, PARTITIONLESS_TEXT)
```
Zopieux: Same remark here on `pangoEscape`.

usb/usb.py
```diff
@@ -0,0 +1,245 @@
...
+        if line:
+            key, val = line.split("=", maxsplit=1)
+            attributeMap[key] = val
+    return attributeMap
+
+def getAttributeMaps(paths):
+    return { path : makeAttributeMap(path) for path in paths}
+
+leaves = getLeafDevicePaths()
+leaves = list(filter(lambda path: not fastIgnore(path), leaves))
```
Zopieux: May I suggest `leaves = [path for path in leaves if not fastIgnore(path)]` with removes both list and filter (and the lambda)?

usb/usb.py
```diff
@@ -0,0 +1,245 @@
...
+attributeMaps = getAttributeMaps(leaves)
+leaves = filter(lambda path: not ignore(path, attributeMaps[path]), leaves)
```
Zopieux: `leaves = (path for path in leaves if not ignore(path, attributeMaps[path]))`

Zopieux:
Thanks for the contribution! The Python script is well formatted and documented. I suggested a few things but it is mainly *aesthetics*, feel free to use them or not.

usb/usb.py
kb100: AFAIK `df -h --output=avail` will never output any character that needs to be escaped ( '&', '<', '>'). Should I guard against this anyway?

usb/usb.py
kb100: As mentioned above `getSpaceAvailable` uses `df -h --output=avail` which I don't believe needs to be escaped. However, I could see how escaping just in case might be wanted. This is where I would do it. Also as far as I know a kernel name will never have bad characters  in it either. You know as I try to explain these things I feel like guarding against bad characters should be done anyway.

Zopieux: I do agree it is very unlikely these strings will ever contain full HTML tags. Nevertheless, even though these functions *should* be safe, the call to `pangoEscape` is basically free.

kb100:
@Zopieux I've addressed all of your concerns in https://github.com/kb100/i3blocks-contrib/commit/f4645ca4b5bfda26e6ef06a1ec5593009e8640a9, what do you think of it now?

kb100:
I've squashed all the commits into one and made the title of the commit match @vivien's suggestions.

Zopieux:
Great! I don't have any further suggestions.

vivien:
Thanks to both of you!

PR #2, opened 2015-08-04
mediaplayer: add support for mpd

thiagowfx:
Also:

- refactoring, making the script more reusable
- mpd support is obtained with `mpc`

Recreated PR from https://github.com/vivien/i3blocks/pull/126.


vivien:
Looks good to me!

PR #4, opened 2015-08-07
Setup initial draft of contributing guidelines

kb100:
This is an initial proposal of what the contributing guidelines should be. It is geared towards being inviting to newcomers. Request for comments.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,1 +1,110 @@
 # How to contribute
+
+So you've written a blocklet for i3blocks and would like to share it with the 
+community, great! Let's just set a few ground rules in order to get your
+blocklet included. 
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", 
+"SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be 
+interpreted as described in [RFC 2119](https://www.ietf.org/rfc/rfc2119.txt).
+
+# Directory guidelines
+
+1. A blocklet MUST be confined to a single directory and the name of the 
+   directory MUST be relevant to the blocklet's purpose or core feature.
+2. A blocklet MUST be human readable.
+3. A blocklet MUST contain a `README.md` explaining what the blocklet does,
+   what its dependencies are, and any special instruction or setup required to
+   use it.
+4. A suggested i3blocks configuration MUST be included in an `i3blocks.conf` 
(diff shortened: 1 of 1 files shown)


Nycroth:
Hey @kb100, I loved your initial direction with the standards. :+1: I went ahead and ported them to the wiki and while doing so make a few changes and add a section to fully guide the process of both oneliners and full blocklets ([Wiki Entry](https://github.com/vivien/i3blocks-contrib/wiki/Contributing)). I'm not sure how we should go about utilizing `CONTRIBUTING.md` and the wiki.

kb100:
Thanks for all the work on the wiki! I may be misunderstanding the purpose of the wiki, but do we really want to have an entire copy of every blocklet's README inside the wiki? Perhaps we should just put pictures and short descriptions in the wiki and link the the corresponding blocklet's README? Or the other way around? Just short description in the README and lengthy explanation in the wiki?

Nycroth:
I'd actually thought about that. :smile: Having the README contain only the head of the wiki (name [linked to wiki], sceenshot(s), description, dependencies, and optional dependencies) would reduce the number of edits it requires. Usually these things only change when the blocklet itself changes (thus requires a PR anyway). The wiki is a very simple edit for everyone. There is no PR or review. With the README name linking to the wiki, finding more is simple. At the same time, details that are often played with (using the block) can be changed many times without too much resistance.

Nycroth:
Probably best for `CONTRIBUTING.md` to contain a small blerb and link to the wiki's contributing guidelines followed by the most current set of Repo Standards. Then we note in the Repo Standards wiki entry that they are only in the Wiki for reference while the official standards live in `CONTRIBUTING.md`. This way there is a little more accounting on the changing of standards.

vivien:
Thanks @kb100 ! Sorry for the delay.

Here's my point on the wiki/source choice: I'd say, put whatever you want or find useful in the wiki, it's there for that. But I'd have a preference for at least clean source. As a user, I tend to trust and browse the source before wiki entries. Also sources are the only source of information if you fetch a tarball a clone the repository. A complete and explicit `CONTRIBUTING.md` file is important, since it is used by Github as a reminder for everyone's PR.

vivien:
Also as @Zopieux noticed in another ticket, PR to update README files or other source of info is not that a big overhead, especially since we have a separated repo and will have several maintainers.

Nycroth:
Fair enough. :+1: I agree with the wiki source split. Anything I can do to help set that up?

PR #8, opened 2015-09-09
batterybar:  displays the battery level as a bar

Keftaa:


batterybar/README.md
```diff
@@ -0,0 +1,31 @@
...
+	
+Left-clicking briefly shows the battery level in percent.<br>
+
+Installation
+------------
+1-Copy the battery_bar script into /usr/share/i3blocks/.<br>
+2-Give it execution permission. (#chmod +x battery_bar).<br>
+3-Add the following blocket to your i3blocks.conf:<br>
+
+[battery_bar]<br>
```
Zopieux: What about using an actual block of code? Use this syntax:

    ```ini
    [battery_bar]
    <snip>
    ```

batterybar/README.md
```diff
@@ -0,0 +1,31 @@
...
+-----------------
+It displays the battery level in a set of five unicode squares (U+25A0).<br>
+It also changes color for more accuracy and to distinguish between<br>
+charging, discharging, charged and AC statuses.<br>
+	
+Left-clicking briefly shows the battery level in percent.<br>
+
+Installation
+------------
+1-Copy the battery_bar script into /usr/share/i3blocks/.<br>
```
Zopieux: Please review the [markdown syntax](https://daringfireball.net/projects/markdown/syntax#list). Ordered lists are formatted with dots:

    1. foo
    2. bar

batterybar/README.md
```diff
@@ -0,0 +1,31 @@
...
+2-Give it execution permission. (#chmod +x battery_bar).<br>
```
Zopieux: Use backquotes for inline code:

    (`chmod +x battery_bar`)

Keftaa: Many thanks, applied all of your notes.

kb100:
I have a few comments about this script before I think it would be ready to merge.

Firstly, the regex does not capture all possible states that acpi reports. In particular, if the script runs right as you plug or unplug your charger, acpi may say unknown. Such an outcome would crash your script. Make sure to check if `$percentage` is not empty first. 

Secondly, the double and triple wildcard matching should really be improved. Why not, e.g. `acpi battery | grep -m1 -o '[0-9]\{1,3\}%$'

Next I suggest that the if blocks and case statements should be styled like the scripts in the main repo. Namely

    if (( ... )); then
        ...
        #four space indentation
    fi

    case "$status" in
        a) color=red ;;
        ...
    esac

   #or
    case "$status" in
        a) 
            color=red 
            ;;
        ...
    esac

    
Also `if (( $BLOCK_BUTTON == 1))` should be `if [ "$BLOCK_BUTTON" -eq 1 ]` as the former will crash when the block button is not defined.

Keftaa:
@kb100 I have updated the script to go with the suggested styling standards, and added the "Unknown" status support. I don't see any need to check whether $percentage is empty.

I used ` if [[ ... ]]` instead of `if [ ...]` because:

«[ or test is the POSIX test command. It can do simple tests on files and strings. In bash, you should use the more powerful [[ instead and ban [ for sake of consistency. [[ can do pattern matching, is faster and safer to use. http://mywiki.wooledge.org/BashGuide/TestsAndConditionals http://mywiki.wooledge.org/BashFAQ/031»

I was also recommended to omit the leading $ in the arithmetic comparisons, because:

«if you write (($foo>2)), there is a risk that foo might contain something like '64#$( rm -fr / )'  or more mundanely 'bar=1' which when the whole command is run, would assign 0 to bar (being the result of 1»

Check the script and feel free to tell me if there any other problems, thank you.

kb100:
Personally, I would like to see the code in between case and esac to be indented as above as well.

The grep docs for `-P` scare me. 

    Interpret PATTERN as a Perl  regular  expression  (PCRE,  see  below).   This  is  highly
                  experimental and grep -P may warn of unimplemented features.

I'm guessing you used it for the 0 length lookahead, but the same can be accomplished simply by deleting the last character:

    acpi battery | grep -m1 -o "[0-9]\{1,3\}%" | tr -d '%'

Note the `-m1` flag is in case the user has multiple batteries..., which would (I believe) cause acpi to output a line for each.

Another suggestion would be to remove the redundant range checks, in pseudocode

    if percentage > 80
        blocks=c1
    else if percentage > 60
        blocks=c2
   ...
    else if percentage > 20
        blocks=c4
    else
        blocks=c5
 
no need to check that percentage is between 60 and 80 in the second case because if it didn't go into the first if block, it must be less than 80, so only check if it is bigger than 60, and similarly for the other clauses. If you feel that your code is more readable doing the redundant checks, my suggestions are definitely not requirements.

Keftaa:
Updated. I'd have liked to implement the multiple battery things but I'm not really sure how to do it, and I believe these cases are pretty rare. I added the `-m1` flag.

I think that the redundant checks are more human-readable, they explicitly look as ranges, which makes it easier to modify in the future in my opinion.

kb100:
In case status is Unknown, does it still report a valid percentage? If not, some null checks are needed. If so it looks good to go.

You may also, if you desire, include any authorship/license/description comments that you desire at the top.

 Squash all your commits into one with `git rebase -i` when you are ready.

Keftaa:
@kb100 Sorry, I got super confused with squashing the commits, fixing the merge conflicts and whatnot. I think the problem was that I was making changes directly on the github website instead of working on local repo then pushing, which I will be doing from now on. 

I tried working with the mergetools which were even more confusing.

I have hence removed my repository and created a new clean pull request which should be ready to implement.

Sorry for the inconvenience.

PR #10, opened 2015-09-11
batterybar: displays the battery level as a bar

Keftaa:


kb100:
No worries about git flubs, everyone makes mistakes learning. One more thing though, your readme instructs the user to put the script in `/usr/share/i3blocks`. The `/usr/share` directory is for files managed by the user's package manager. For user managed files, one should use `/usr/local/share`. I'm not sure if i3blocks will automatically look in that directory for scripts though (test and let us know?).

What I do in my blocks is to add a `command=$SCRIPT_DIR/batterybar`, some others just write `command=./batterybar` but I think this is a little more confusing.

Keftaa:
Alright, done. I have also successfully run `git rebase -i` with no errors, but I still see "2 commits" here. Am I missing anything ?

kb100:
When you rebased, did you squash the second commit? So locally the commit history has only one additional commit? After than you `git push -f` to force push to your repository (must use `-f` because squashing a commit changes history).

Keftaa:
Okay so I removed a whitespace just to commit again, seems good now. The weird thing is that I had to run `git rebase -i HEAD~[number of commits+1]` for it to work.

Keftaa:
UPDATE: the end user can now optionally specify their own set of colors.

kb100:
Up to @vivien if he wants to merge now.

vivien:
Thanks @Keftaa for the script and @Zopieux/@kb100 for the review. I updated the commit message and removed the HTML tags from the readme, since they don't add much values.

PR #12, opened 2015-10-02
apt-upgrades, usb: minor maintenance and new config options

kb100:
Two commits, one for each blocklet.

### DIFF
--- a/apt-upgrades/README.md
+++ b/apt-upgrades/README.md
@@ -13,11 +13,12 @@ Suggested: fonts-font-awesome
 # Options
 
 ```
-Usage: apt-upgrades [-s refresh_symbol] [-o] [-c refresh_color] [-n nonrefresh_color] [-h]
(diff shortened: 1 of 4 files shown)


Nycroth:
USB looks fine. I can't check apt-upgrades though. I use pacman instead.

kb100:
Does @vivien use apt? I hope I'm not the only user of my blocklets, though I may be the only one actively checking pull requests...

vivien:
On Arch too... Looks fine to me anyway.

jpleau:
@kb100 I actually have been using two of your blocklets: apt-upgrades and usb. They are both working great on my desktop (running Debian sid) :)

Sorry for the lack of feedback, I have been busy with lots of stuff, should be back to normal once winter starts !

Nycroth:
No worries @jpleau. :+1:

PR #13, opened 2015-10-05
example_blocklet: Add template directory and README.md

Nycroth:
I think a lone README.md is all that is needed. Feel free to suggest changes.

Edit: Hmmm... I may have been out of the loop too long. I checked the `CONTRIBUTING.md`, a template directory and README may no longer be wanted. I'm cool either way. Its here if you do.

### DIFF
--- a/example_blocklet/README.md
+++ b/example_blocklet/README.md
@@ -1,0 +1,71 @@
+# [Script Name](https://github.com/vivien/i3blocks-contrib/tree/master/script_name)
+
+*Screenshot(s) (Cropped), minimum of one. Delete this italic text when posting.*
+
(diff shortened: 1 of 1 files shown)


jpleau:
Very nice !

I think it'd be great use it as a template, and link to it from https://github.com/vivien/i3blocks-contrib/blob/master/CONTRIBUTING.md (a link on the README.md word, third bullet for guidelines)

We could rename this file so it's not README.md, but something that would indicate a template / skeleton.

A few suggestions (I'm sure someone can come up with better idea than I do for this)
 * README.skel.md
 * README.template.md
 * README.tpl.md

What do you think?

Nycroth:
I like that `template` is clear but I think `skel` is a short yet understandable extension. I'd probably go `template` though.

Nycroth:
Creating a real example blocklet.

PR #15, opened 2015-10-06
Add LICENSE.md to repo

Nycroth:
Noted the license wasn't in the repo, I've just copied it from the wiki. I assume the project is also GPLv3 like i3blocks.

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,4 +1,19 @@
-All blocklets are licenced individually. Consult the `LICENCE.md` contained within each blocklet's directory for the licencing under which the contained blocklet code is published. All blocklets without a `LICENCE.md` in their directory, and all code within the parent of this repository are published under the GNU General Public License.
+This project is a repository project composed of many individual blocklet
+projects. As such, we empower developers to license their blocklet source as
+they see fit with one major caveat.
+
+This repository servers as the single packaging source for most OS
+distributions. As such, we insist that any license you choose meet both the
+[Debian](http://www.debian.org/social_contract#guidelines) and
+[Fedora](https://fedoraproject.org/wiki/Licensing:Main?rd=Licensing#License_of_Fedora_SPEC_Files)
+free license specifications. For a handy chart, see
+[Wikipedia](https://en.wikipedia.org/wiki/Comparison_of_free_and_open-source_software_licenses#Approvals).
+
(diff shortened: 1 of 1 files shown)


jpleau:
Question: should we allow blocklets with licenses that are not compatible with GPLv3 ? 

ie: http://www.gnu.org/licenses/license-list.en.html#GPLIncompatibleLicenses

Nycroth:
I've no horse. I personally don't believe in intelectual property.

This is something Vivien will have decide.

kb100:
I think any free software license should be acceptable. Let the content creator decide how to license his or her code. As long as someone else's license doesn't negatively affect me, I don't have any issue with my code being in the same repository as theirs. Since we allow multiple blocklets that accomplish the same task, in the event of a true conflict of interest in deciding the license of a blocklet, the blocklet can simply be forked or rewritten under another license.

That said, if in the future we decide that i3blocks-contrib should be packaged, e.g., for debian, then we will need to be stricter about what licenses are allowed, or the maintainer will have to exclude those blocklets do not conform to the distro's licensing standards. The maintainer could also package conforming packages into i3blocks-contrib and nonconforming packages into i3blocks-nonfree.

jpleau:
@kb100 That was the reason why I was mentionning this ;)

I planned to package i3blocks-contrib soon  [1] for Debian / Ubuntu, I
don't think licenses will be an issue however. Mixing GPL-licensed scripts
with other BSD-licensed scripts for example wouldn't cause problems,
there's no linkage involved. I would like to avoid having to package a
non-free version of i3blocks-nonfree, for simplicity's sake

1. Most likely when i3blocks-1.5 and i3blocks-contrib-1.5 are released :)

2015-10-06 18:15 GMT-04:00 kb100 <notifications@github.com>:

vivien:
Same as @jpleau. I don't mind multiple licenses, but:
1. I would strongly recommend free (as in **GPL**) software ;-)
2. I want to make it simple for packagers.

Nycroth:
Just changed the license intro. If this change is accepted, we had best update the `CONTRIBUTING.md` guidelines.

Nycroth:
Is anyone opposed to merging this?

kb100:
Looks good to me, squash and merge.

Nycroth:
Merged.

PR #16, opened 2015-11-02
shutdown_menu: add new script providing menu for easy reboot, shutdown etc.

bchretien:
As explained in the README, this provides a simple shutdown menu thanks to dmenu/rofi and systemd. This can certainly be improved, but I figured this may already be useful for other i3blocks users.


kb100:
Looks like a good idea, i'll review this carefully soon.

kb100:
Would it be possible to make the menu clickable? It seems that the only purpose the blocklet serves is to create a quit button for the user to click. But the menu that pops up is not clickable, so one has to go to the keyboard to select an option. But if one was already at the keyboard, this defeats the purpose of having a button to click, doesn't it? Why not just call the script from a keyboard shortcut or the command line?

bchretien:
The menu is clickable if you use rofi. I guess we could use zenity in replacement of dmenu to have both options clickable.

:+1: for selection warnings, I'll give it some thoughts.

bchretien:
These new commits should solve the problems you raised:

- both rofi and zenity are supported (and both support mouse interactions),
- a `-c` option was added, it triggers a confirmation popup for all actions except cancel and lock.

kb100:
The default behavior is up to you, but could you add a command line switch for "prefer rofi" and "prefer zenity" so that users who have both installed can still use the one that isn't the default.

bchretien:
Users can now do `shutdown_menu -p rofi` or `shutdown_menu -p zenity`.

kb100:
When I use your script "lock" my VM becomes unresponsive, and putting in my password doesn't give any feedback, it just sits frozen as if the keyboard was unplugged, and I must kill the VM. Directly calling `i3lock -c "#111111"` does not have this problem.

bchretien:
Hmmm that's weird, it does work on my system... The doc specifies `--color=rrggbb` for i3lock 2.7, so I changed the line in the menu to:

```sh
  [Lock]="${LOCKSCRIPT:-i3lock --color=${BG_COLOR#"#"}}"
```

Could you tell me if that works for you?

kb100:
That doesn't seem to be the issue, even if I hard code in the color the problem persists.

kb100:
I figured out how to fix it, instead of calling i3lock directly, use i3-msg

    i3-msg exec i3lock -c #111111

I imagine all of the things should be called using `i3-msg` in this manner, so that the script is allowed to exit nicely.

bchretien:
Ok, I still changed the i3lock command (the doc explicitly says `rrggbb` is expected, so I omitted `#`), and now all commands will be run with `i3-msg -q exec`.

kb100:
You may or may not want to recommend using the FontAwesome power-off icon for the label.
http://fontawesome.io/icon/power-off/

At this point we are about ready to merge. Can you squash all the commits into one?

bchretien:
Rebased and added a word on FontAwesome.

jpleau:
Looks good to me, thanks for the script !

kb100:
Merged and thanks for the contribution.

bchretien:
Well thanks for the feedback :)

PR #17, opened 2015-11-14
bandwidth: add script to moitor bandwidth usage

kb100:


Nycroth:
Works as expected. However, might I suggest using $BLOCK_INSTANCE to pass the desired interface?

kb100:
@Nycroth I set $BLOCK_INSTANCE to the default interface, falling back to eth0 if it is empty. The user can still use the `-i` flag to specify the interface if they wish.

Nycroth:
Looks great! I'm happy to merge this. I'll give until Monday morning for others comments and reviews and then I'll merge it.

kb100:
I believe @vivien was having an issue and I'de like to get that resolved before merging.

vivien:
Arf I cannot find my comments. I post it here anyway. I get this (interface doesn't matter):
````
$ ./bandwidth/bandwidth
awk: cmd. line:2: BEGIN{old_received=;old_sent=}
awk: cmd. line:2:                    ^ syntax error
awk: cmd. line:2: BEGIN{old_received=;old_sent=}
awk: cmd. line:2:                              ^ syntax error
````

kb100:
@vivien Are you using the newest commit? What does your `/proc/net/dev` look like?

vivien:
Indeed, on my machine (4.2.5-1-ARCH #1 SMP PREEMPT x86_64 GNU/Linux) the /proc/net/dev file shows looks like this:

````
$ cat /proc/net/dev 
Inter-|   Receive                                                |  Transmit
 face |bytes    packets errs drop fifo frame compressed multicast|bytes    packets errs drop fifo colls carrier compressed
enp0s25: 42133711   84234    0  620    0     0          0     12620  5264266   31890    0    0    0     0       0          0
    lo:   24312     255    0    0    0     0          0         0    24312     255    0    0    0     0       0          0
wlp3s0:       0       0    0    0    0     0          0         0        0       0    0    0    0     0       0          0
````

And this is fixed with your new commit, thanks!

Can we default to something like `ip route | awk '/^default/ { print $5 ; exit }'` instead of `eth0` for the interface? I have `enp0s25` and `wlp3s0` and I'm often switching between them.

kb100:
@vivien  yes, I'll make the chain of command: ip route default, overwritten by BLOCK_INSTANCE, overwritten by `-i iface`.

kb100:
@vivien check it now. I don't know about switching automatic switching of interfaces. In its current form if you switch interfaces then you will have to restart your bar in order for it to notice the change in default. I'm not sure if dynamic switching of interfaces can be done without adding a significant amount to the code.

vivien:
Picking a default interface on startup is nice and convenient, but beside that, we should avoid doing magic. An explicit `i3-msg restart` when you change your setup is simpler ;-)

I tested the latest commit and it works great! Please fix the 2 typos: *moitor* (in the commit title) and *blocket* (in the script) and the change is good to merge for me!

kb100:
@vivien oops, how embarrassing. Fixed.

Nycroth:
Merged! Thanks @kb100.

vivien:
That is a good example for the `persist` interval that I wanted to hack for a long time. Thanks for doing it @kb100!

PR #24, opened 2015-12-29
mediaplayer spotifix

roobre:
Some players (like spotify) have a little delay between changing the current song and exposing the new song's data to DBUS. Waiting an small amount of time (10ms) between issuing the `next` command and retrieving the current song metadata fixed this.

This uses `Time::HiRes`, but it's an standard perl library since 5.8, so it shouldn't break anything.

The `buttons` subroutine call was also moved before the `metadata` call in order to allow the new data to be retrieved.

### DIFF
--- a/iface-ng/README.md
+++ b/iface-ng/README.md
@@ -1,0 +1,25 @@
+# iface-ng
+
+Display information about network interfaces, more flexible than default `iface`.
+
+``` ini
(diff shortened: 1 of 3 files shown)


roobre:
Sorry, I accidentally included the commits from my previous pull request into this :(

Closing and fixing.

PR #25, opened 2015-12-29
mediaplayer fix for slow players, fixed a bug which caused (null) - (null) being displayed

roobre:
Some players (like spotify) have a little delay between changing the current song and exposing the new song's data to DBUS. Waiting an small amount of time (10ms) between issuing the `next` command and retrieving the current song metadata fixed this.

The `buttons` subroutine call was also moved before the `metadata` call in order to allow the new data to be retrieved.

Please note this uses `Time::HiRes`, but it's an standard perl library since 5.8, so it shouldn't break anything.

### DIFF
--- a/mediaplayer/mediaplayer
+++ b/mediaplayer/mediaplayer
@@ -22,7 +22,11 @@
 # (playerctl will attempt to connect to org.mpris.MediaPlayer2.[NAME] on your
 # DBus session).
 
+use Time::HiRes qw(usleep);
 use Env qw(BLOCK_INSTANCE);
+
+use constant DELAY => 50; # Delay in ms to let network-based players (spotify) reflect new data.
+use constant SPOTIFY_STR => 'spotify';
 
 my @metadata = ();
 my $player_arg = "";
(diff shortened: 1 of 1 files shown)


kb100:
A few comments/questions:

-  Commits MUST begin with the name of the blocklet in question and SHOULD be written in imperative form, e.g. `mediaplayer: 	add 10ms sleep to let players reflect new data`

- Where does this 10ms number come from? Will that be enough wait time for users with slower machines?

- A delay isn't necessary for all media players, so the code should only pause if you are using a player that needs it.

roobre:
Hm, sorry about the commit name, I thought it was the pull request name the one it should begin with the script name.

It might be a way to receive notifications via DBUS, but definetly not through `playerctl`. And I think we should either use `playerctl` only, or go directly against DBUS (which may be a bit tricky imo).

You're right, only players which rely on network (namely spotify) suffer this issue. I'll add a check for that in a moment.

roobre:
Ok, I've increased the delay a bit, added it as a constant, and check delay only if "spotify" is specified.

Also, I've spotted a bug that caused `(null) - (null)` to be output if `$BLOCK_INSTANCE` was defined, but the player were not running. This was (again) `playerctl`'s fault, as `playerctl --player='foobar' metadata artist` returns 0 even if no player named "foobar" is running.
Anyway, this was also fixed by checking the output is not `(null)`.

mediaplayer/mediaplayer
```diff
@@ -88,16 +94,17 @@ sub mpd {
...
-    exit(0) if $?;
-
-    buttons;
+    exit(0) if $? || $artist eq '(null)';
 
     push(@metadata, $artist) if $artist;
 
     my $title = qx(playerctl $player_arg metadata title);
-    exit(0) if $?;
+    exit(0) if $? || $artist eq '(null)';
```
kb100: :106s/artist/title/

roobre: Definetly. My bad, fixing... ^^U

kb100:
Everything looks good to me. I'll let this sit for a day or two to let other contributors comment if they want. Please squash your commits into one in preparation for the merge.

kb100:
@roobre I'm ready to merge if you squash your commits into one

roobre:
Sorry, I've been busy the last week. I'll squash and push them tomorrow.

roobre:
Commits squashed!

PR #30, opened 2016-01-27
usb and apt-upgrades maintenance

kb100:
I've added command line configuration for the usb script and used `[[` instead of `[` in the apt-upgrades script.

### DIFF
--- a/apt-upgrades/apt-upgrades
+++ b/apt-upgrades/apt-upgrades
@@ -45,9 +45,9 @@ aptitude full-upgrade --simulate --assume-yes |\
     grep '[0-9]\+' |\
     xargs echo)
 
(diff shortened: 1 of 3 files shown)


jpleau:
@kb100 usb changes are cool, thanks!

For apt-upgrades, I notice (and it's the same for a lot of scripts), that they heavily depend on 1) bash is installed and 2) it's in /bin/bash

For example on FreeBSD, /bin/bash probably does not exist (bash would be in /usr/local/bin/bash). So using `#!/usr/bin/env bash` shebang would work there.

Unless we are targeting GNU/Linux only, do you think we should try to make shell scripts portable?

kb100:
I had no idea that's why people use `env bash` instead of just bash. Yes I don't see any harm in being more portable. That's a topic for a different pull request though. It should be straightforward though.

jpleau:
@kb100 fair enough ! I'll merge this one!

Issue #28, opened 2016-01-12
batterybar: Bad output if battery unplugged


kb100:
@lathan 

With my battery unplugged (laptop plugged into charger):

    $ ./batterybar               
    <span foreground="">■</span>
    $ BLOCK_BUTTON=1 ./batterybar
     <span foreground="">%</span>
    <span foreground="">■</span>

My `acpi battery` output is empty. Also note the extra space at the beginning of line 4.

lathan:
Hi @kb100 . Thnx. I'll try to look into this tomorrow from my work laptop. My macbook doesn't respect my freedom enough to allow me to remove the battery to see what acpi returns.

lathan:
It turns out, both my laptops are a bit weird in this area.

@kb100 could you tell me what the full output of the following commands is on your system when you have unplugged the battery and connected the charger?

```
acpi
acpi battery
```

kb100:
Both `acpi` and `acpi battery` produce no output whatsoever when my battery is unplugged.

PR #31, opened 2016-01-27
Batterybar fix, handle empty ACPI output

kb100:
fixes https://github.com/vivien/i3blocks-contrib/issues/28

### DIFF
--- a/batterybar/batterybar
+++ b/batterybar/batterybar
@@ -22,9 +22,9 @@
 
 
 
-output=$(acpi battery)
+output=$(acpi battery 2>/dev/null)
 percentage=$(echo "$output" | grep -o -m1 '[0-9]\{1,3\}%' | tr -d '%')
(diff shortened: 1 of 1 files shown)


kb100:
Any idea why it is showing that previous commit? My commit is only the batterybar one.

kb100:
There we go, I just needed to rebase onto the upstream...

jpleau:
@kb100 

If I run this on a desktop, is this the expected behavior? I don't have a laptop or anything with a battery this is all I can test :(

http://i.imgur.com/SPk1GfA.png (the red part on the right)

kb100:
AFAIK yes that is what the author intended. it's not the most pleasing looking IMO.

kb100:
Can the author @lathan comment?

lathan:
Sorry for the late response.
* I'm not the original author.
* I made some changes recently but never tested on desktop or generally without acpi installed.

lathan:
I don't have any suggestions beyond this.

PR #33, opened 2016-02-13
key_layout optimization

METAJIJI:
Minimizing the pipes, and system calls. There was only awk and no bash.

### DIFF
--- a/key_layout/key_layout
+++ b/key_layout/key_layout
@@ -1,6 +1,7 @@
 #!/usr/bin/env bash
 
 # Copyright 2016 Patrick Haun
+# Edited: Denis Kadyshev
 #
 # This program is free software: you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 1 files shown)


kb100:
Hi, would it be possible to measure performance increase? Like run the block a large number of times each way and compute the average execution time. (I don't know exactly what I'm asking for here so it's a learning process for me too) Also, I think your `if, if` blocks might benefit from being `if, else if`.

METAJIJI:
Exaple simple tests:
```
$ time bash new.sh 
us,ru,
real	0m0.011s
user	0m0.004s
sys	0m0.005s

$ time bash old.sh 
us,ru ,

real	0m0.017s
user	0m0.009s
sys	0m0.010s
```
`new.sh` is faster, because `setxkbmap` run only once and run only alone `awk`.
`old.sh` is slow, because `setxkbmap` and awk run at 2 times and `tr`.

Also `new.sh` fix the trailing space after `us,ru`.

Added performance is very small, but it is more the right way.

kb100:
I like the idea but the code looks a little clumsy. What do you think about

    setxkbmap -query | awk '
        BEGIN{layout="";variant=""}
        /^layout/{layout=$2}
        /^variant/{variant=$2}
        END{printf("%s%s",layout,variant)}
    '

METAJIJI:
Yes, this is good variant! I'm create new commit.

kb100:
Ready to merge, please squash the two commits into one and make the commit title of the form

    key_layout: optimize by calling programs only once

METAJIJI:
Done!

kb100:
Thanks for your contribution!

PR #39, opened 2016-03-23
keyindicator: interpret LED mask as hex instead of dec

kb100:
Previously, the block would give no output if a user's LED mask
was (e.g.): ffffe7fc

Use `xset -q | grep LED` to see your current LED mask

This commit fixes that issue by correctly interpreting the mask as hex
instead of decimal.


kb100:
@vivien Perhaps you can review this commit?

PR #42, opened 2016-03-25
Removed endline ($) to make cmus work

cheukyin699:
Because of how $player_arg is defined, there is a single quote between
$BLOCK_INSTANCE and endline ($).

```perl
$player_arg = "--player='cmus'";
```

### DIFF
--- a/mediaplayer/mediaplayer
+++ b/mediaplayer/mediaplayer
@@ -113,7 +113,7 @@ sub playerctl {
 if ($player_arg eq '' or $player_arg eq 'mpd') {
     mpd;
 }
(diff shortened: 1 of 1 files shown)


kb100:
Hi thanks for noticing and fixing this issue. Please edit your commit message to begin with `mediaplayer: ` and I will merge.

Also, it seems to me that the script will fail if you put `instance=mpd` in your config. If you are up for it, this is probably a quick fix as well.

cheukyin699:
Thanks! Commit message fixed as well as mpd (I think).

PR #49, opened 2016-05-29
Fix megabyte display in bandwidth blocklet

dshoreman:
MBytes are wrongly displayed as Mbits. This PR removes the division by 8 so that it prints MB as MB.


kb100:
Good catch

PR #50, opened 2016-06-01
brightness : increase or decrease your screen brightness

thomasmerel:


kb100:
What are the lines:

     dir=$(dirname $0)
    source $dir/util.sh

for?

Also, xbacklight does not work with all screens, e.g. mine, so would it be possible to handle and document the case when xbacklight returns "No outputs have backlight property". Perhaps fall back to using `xrandr --output NAME --brightness VAL` for each of the active outputs.

PR #55, opened 2016-08-24
Add email blocklet.

bergentroll:


bergentroll:
Sorry, I forgot to close connection. I make commit tomorrow.

bergentroll:
Alright, done!

kb100:
This looks promising. A few comments:


1.  Include a clear Dependencies/Requirements section devoted to only the dependencies, see [this block](https://github.com/vivien/i3blocks-contrib/tree/master/monitor_manager) for a good minimal example.

2.  The repository-wide convention is to assume that the script is downloaded to `$SCRIPT_DIR`. No explanation of this is needed, but a short one may be included.

3. Avoid reference to external blocks when possible, I don't think `dimmer` is required for the `email` blocklet.

4. I believe `keyring` is the same as in the debian package `python3-keyring` and `python3-keyring.alt`. If so, it would be preferred to mention these packages and give the pip install only as an alternative to your distro's official repository.

5. Since the block only deals with imap would you consider renaming blocklet to `email-imap`?

bergentroll:
I changed README.md according to you message.
IMAP is most common email protocol today I believe. Also if I'll mark it in the block name, It could be reasonable to specify also kind of IMAP (TLS/SSL). But I'll make as you insist.

kb100:
Some more comments:

1. It is very bad practice to put a plaintext password in a (world readable) file. Please make a note that setting the $PASS variable manually is highly discouraged and should be for debugging purposes only. Keyrings are the way to go.
2.  On a multi-user system, it may not be possible (or desirable) for a user to edit their settings in the system email script, not really sure if there is anything you can do about this at this point since adding command line arguments would be an atrocity.
3. Give the README a quick proofread for grammar and spelling.

Once you feel the block is ready to merge, please squash all your commits into one, see https://github.com/vivien/i3blocks-contrib/blob/master/CONTRIBUTING.md for detailed instructions.

bergentroll:
Hi!
Please, check new version.

kb100:
Hi, it looks like you replaced the root README of the repository, please correct this. Also, The get_PASS function doesn't use or need the PASS argument, please remove. Finally, len(config['PASS']) will error if config['PASS'] is None, you can just write `if not config['PASS']):` instead.

After these corrections I think I will be ready to merge

kb100:
Oh also, `/usr/bin/env python3`, otherwise block will not be able to import `configparser`

bergentroll:
Right, I don't know how I broke the /README.md.
Fixed now.

kb100:
Merged, thanks for your contribution.

bergentroll:
Thank you!

PR #65, opened 2016-11-23
Add vol up and vol down in mpd

hillz1:
Mouse scroll up = song volume increased by 10
Mouse scroll down = song volume decreased by 10


kb100:
Merged in https://github.com/vivien/i3blocks-contrib/commit/0d070252dfac15ae062098567227b5c9f26fb5a5

Thanks for this!

PR #68, opened 2017-01-05
keyindicator: add --hide option to not show anything if switch is off.

wizpig64:
I wanted this script block to show nothing, rather than a grayed out CAPS/NUM, when the respective switches were off. This PR adds a --hide option which does that.

I've never worked in perl before but hopefully it all looks right.

Thanks!


kb100:
Thanks for the improvement and happy perling.

PR #72, opened 2017-01-21
Use /usr/bin/env perl instead of /usr/bin/perl for portability

alibabzo:


kb100:
Thanks for catching this, merged.

Issue #76, opened 2017-01-31
Does anyone interested in yet another kbd layout script based on xkblayout-state?


qemi:
key_layout doesn't show currently active keyboard layout; instead, I get a list of all enabled layouts: like us, it

kbdd_layout uses kbdd for the same, but I don't want to keep another process running for such small purpose  (at least this process stores some state in memory, but mentioned script is just a stateless query tool).

There's a small program https://github.com/nonpop/xkblayout-state
For Arch Linux users, it's available in AUR, others can simply clone repo and make it manually.

Usage:
$ xkblayout-state print %s
us

kb100:
I agree that the keyboard layout scripts here are not ideal, so I would consider adding another. However, I should give two warnings:

1. I personally review blocklets before merging. If your blocklet is 5 files of C++ it may take a while. Single file short scripts are preferred.

2. Key layout can change without user interaction, or be set on a per-window basis. the solution using kbdd uses a subscription method so it updates automatically whenever it receives information about updated keyboard layout. If you don't have a daemon subscribed to get keyboard layout info, then (it seems to me that) the only thing you can do is poll the layout frequently or set a symbind to signal i3blocks whenever you manually execute a layout switch command, but you will not be able to handle per-window layouts.

If you find a novel solution I am always willing to read it though.

PR #77, opened 2017-02-01
key_layout2 script added

qemi:
Solves the [#76](https://github.com/vivien/i3blocks-contrib/issues/76).


kb100:
xkblayout-state segfaults for me on every run, I believe this is the same bug as https://github.com/nonpop/xkblayout-state/issues/4, so I can't test this block until that issue is resolved.

qemi:
Thanks @kb100, let's wait for fix. I'll look on this later and maybe create PR to that repo.

PR #79, opened 2017-02-08
Catch up with the main repo

roobre:


roobre:
Yikes, sorry, I created the pull request in the wrong direciton.

PR #80, opened 2017-02-08
wlan-dbm: Added wlan-dbm script

roobre:
This is a small script I wrote to show the wireless interface signal strength in dBm instead of percent, as it is more meaningful to me.


kb100:
I think some instructions are missing. Doesn't one need root to use `iw`? i3blocks will not be running as root and therefore will not be able to execute `wlan-dbm` properly.

roobre:
It does for commands which manipulate the state of the card. Just querying it (like `iw list` or `iw link`) does not require root privileges.

kb100:
I just tested and it works directly calling `/sbin/iw` on debian. Handle this and I will be happy to merge. Would it also be possible to add a flag to give the signal strength percentage for those users who prefer it over dBm? (If so a name change to wifi-signal-strength might be in order)

roobre:
Hello,

Yes, I just confirmed it yesterday too, so I've added a fallbeck to /sbin/iw if it is not found in the user's path.

I'll look into the flag thing and see if I can do it without adding too much complexity.

kb100:
See https://stackoverflow.com/questions/15797920/how-to-convert-wifi-signal-strength-from-quality-percent-to-rssi-dbm for conversion to percent, it is very simple. Also adding command line flags is a breeze with getopts, see https://github.com/vivien/i3blocks-contrib/blob/master/apt-upgrades/apt-upgrades for a short example

roobre:
While dBm is an absolute measure of the signal strength, "signal percent" it's just an approximate measure relative to which dBm values the code consider "optimal" and "insufficient". These values, however, depend heavily on the card and antenna being used, but are rarely specified by the manufacturer. Hence they can only be guessed (as far as I know) via experimentation.

I'll implement this conversion with some default values based on my experience and some research, and allow customization of them via command line arguments. Keep in mind that using the defaults will provide an arbitrary approximation, and some cards may work flawlessly at 40%, while others will consider it insufficient.

Stay tuned.

PR #82, opened 2017-02-19
Keyindicator: only set bg colors if specified

jolange:
- Small option syntax fix in README
- Only setting background colors if explicitly specified. For users with a non-black i3bar this avoids the need to specify the background colors.


kb100:
Thanks for this, merged.

PR #84, opened 2017-02-26
key_layout: Fix detect layout on multilingual

METAJIJI:
Change detect method from `setxkbmap` to `xset -q`, because `setxkbmap`  can not detect when more than one keyboard layouts.

### DIFF
--- a/key_layout/key_layout
+++ b/key_layout/key_layout
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
 # Copyright 2016 Patrick Haun
-# Edited: Denis Kadyshev
+# Edited: 2017 Denis Kadyshev
 #
 # This program is free software: you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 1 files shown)


METAJIJI:
But previous code show only `us,ru,winkeys`, and this not informative.
I think display the keyboard layout is much more useful.

And new code is show `us` or `ru` or `other` and this more informative also i never seen `winkeys` and similar in regular OS (win, mac, linux, etc), may be drop this feature? How use `winkeys`?

jolange:
> But previous code show only us,ru,winkeys, and this not informative.

I don't see what you mean. The current script shows whatever setxkbmap returns (in my case `de`), which exacly what it should do, right?

> And new code is show us or ru or other

In my case it shows `en` instead of `de`.

METAJIJI:
First set multiple keyboard layouts and set switch shortcut Alt+Shift (in my case):

`setxkbmap "us,ru" ",winkeys" "grp:alt_shift_toggle"`

Then run script and see:
```
$ ./key_layout
us,ru,winkeys
```

Switch between layouts (Press Alt+Shift) does not change the output `key_layout`.

jolange:
Ah, I see your problem. I did not know that toggle mechanism.

Still, your code does not work generically, see also the comments [here](http://unix.stackexchange.com/questions/12072/how-do-i-get-current-keyboard-layout/27688#27688).

METAJIJI:
I see this thread and write awk function for workaround this and now my code is work with caps and num lock if you mean this point :)

jolange:
Maybe this might be interesting for you: https://github.com/nonpop/xkblayout-state

kb100:
Keyboard layout is, for some reason, one of those things that has not been properly unified in linux yet. As such, "getting the current layout" is undefined and many programs do it in different ways and give different results. In this repo, we have two blocks, key_layout, and kbdd_layout, and possibly another in https://github.com/vivien/i3blocks-contrib/pull/77 soon. If you are toggling using layout groups, then use kbdd_layout, which handles this properly.

PR #86, opened 2017-04-02
key_layout: put possible variant in parentheses and separate with space

jolange:
e.g. output "de (nodeadkeys)" instead of "denodeadkeys"


kb100:
Thanks for this!

PR #87, opened 2017-04-04
[volume-pulseaudio] Fix a bug which prevents users from setting custom icon for muted state

wh11e:
[This line](https://github.com/vivien/i3blocks-contrib/blob/597a27df27/volume-pulseaudio/volume-pulseaudio#L31) is never executed because [the X was missing](https://github.com/vivien/i3blocks-contrib/pull/87/files) so I fixed it


kb100:
Nice catch, thanks.

PR #88, opened 2017-04-26
added support for multiple batteries

fluktuid:
haven't seen multiple battery support for the batterybar, so I added it.

PR #89, opened 2017-04-27
Multi battery support for batterybar

snowiow:
Hi, 
I had a problem with batterbyar, that it only showed the first battery of the acpi battery output. Because the first one is my backup battery I couldn't see the actual battery status of the battery which is used in the first place. I made some changes, that a battery bar for every battery is shown. I tried to not change the functionality of the initial script and wanted to just extend the script in case there is more than one battery. I don't know if this should update the existing batterybar script or should be an additional script, but I can change it accordingly if it's wanted. 

Also: I'm a pretty big bash scripting noob and I don't know if the stuff I've written is common best practice. So if anyone has some tips on how to improve the given code, I'm open to change it.

Greetings

### DIFF
--- a/batterybar/batterybar
+++ b/batterybar/batterybar
@@ -88,7 +88,7 @@ do
     fi
 
     if [[ "${statuses[$i]}" = "Unknown" ]]; then
-        squares="<sup>?</sup>${squares[$i]}"
+        squares="<sup>?</sup>$squares"
     fi
 
     case "${statuses[$i]}" in
--- a/batterybar/batterybar
+++ b/batterybar/batterybar
@@ -22,7 +22,6 @@
 readarray -t output <<< $(acpi battery)
 battery_count=${#output[@]}
 
-remainings=()
 for line in "${output[@]}";
 do
     percentages+=($(echo "$line" | grep -o -m1 '[0-9]\{1,3\}%' | tr -d '%'))
--- a/batterybar/batterybar
+++ b/batterybar/batterybar
@@ -19,16 +19,23 @@
 #  MA 02110-1301, USA.
 #  
 #  
+readarray -t output <<< $(acpi battery)
+battery_count=${#output[@]}
 
+remainings=()
+for line in "${output[@]}";
+do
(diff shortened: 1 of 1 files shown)


fluktuid:
isn't this code always showing the status color of bat0?

snowiow:
You mean the `$color` variable? no it will be overwritten every iteration. at the end of each iteration I append the current colored bar. After I iterated over every battery, everything will be printed at once.

fluktuid:
no, i mean
```
if [[ "${statuses[$i]}" = "Unknown" ]]; then
        squares="<sup>?</sup>${squares[$i]}"
```

because on my computer the not used battery has the status 'unknown'
so I think
`squares="<sup>?</sup>$squares"`

see the code at #88

snowiow:
yes, you are right. Didn't fixed it from an earlier version, thanks. Didn't saw your PR yesterday, if you already fixed the multi battery problem, I can close this one right?

fluktuid:
Hopefully it's fixed now :crossed_fingers:

kb100:
@snowiow After reviewing both yours and @fluktuid's pull request, I believe yours it better written and more general. Thank you both for your involvement. If you could both test this pull request, then I think I would like to merge this one.

fluktuid:
@kb100 If  so, I will close my request.

fluktuid:
Then I'll make you a bit work
- It is desired that the bar becomes shorter?  bb

snowiow:
for me yes, but if you guys like it better with the empty boxes, we could change it. Alternatively we could add an option in  that one can choose either way (doesn't need to be part of this pr)

kb100:
I don't think it is necessary to add an empty boxes option in this pr, maybe another one if you feel like it. Currently a user can still set the `min-width` option if they want to prevent their bar from shifting.

kb100:
I merged this, making only a minor change (replacing ` .. ` with $( .. )). Thanks and if you would like to work on this more, you may want to gradefully handle what happens to the block if no battery is plugged (some laptops can have battery unplugged while plugged to AC). Currently the behavior is

    No support for device type: power_supply
    <span foreground="">■</span>

snowiow:
I could try, but unfortunately I can't test it, because one of my batteries is soldered.

PR #93, opened 2017-05-22
bandwidth2: add space before byte unit, enforce all consistent widget character length

matthiaskrgr:
"B/s" was only 3 chars long while "KB/s" and the others were 4 chars long. This would cause the block widget to change length all the time and was shifting the entire content of the i3bar back and forth which was very annoying.


kb100:
Looks good, thanks.

PR #95, opened 2017-05-29
volume-pulseaudio: add option -m to move all input to new default sink

jerdna-regeiz:
Currently playing audio streams do not switch automatically to new default sinks.
To move all pulse inputs when changing the default default sink (per click), I added a small function and a new new option -m

Cheers, Andrej


kb100:
I had no idea this was even possible, great find. I actually think most people would want this to be default behavior (no need to make it an option). What do you think?

PR #99, opened 2017-06-22
arch-update: add --hide option

jolange:
option `-H/--hide`: produces no output if no updates available

@snowiow: Since this is completely your work up to now, you might want to add your two cents about this change.

### DIFF
--- a/arch-update/arch-update.py
+++ b/arch-update/arch-update.py
@@ -29,6 +29,12 @@ def create_argparse():
         action='store_true',
         help='Include AUR packages. Attn: Yaourt must be installed'
     )
+    parser.add_argument(
+        '-q',
+        '--quiet',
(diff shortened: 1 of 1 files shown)


snowiow:
Hi, thanks for your contribution. I would call the option quiet. Then you can use -q as the short argument. That way it would be more distinct from the help option. Beside of that it looks good to me.

jolange:
> I would call the option quiet.

Also fine with me, changed.

kb100:
Thanks for this.

jolange:
Wow, that was fast ;-)

kb100:
Commit was very straightforward so it was easy to review :) Also I added the quiet option to the README

jolange:
Ah thanks, I forgot that!

PR #100, opened 2017-07-03
mediaplayer: add rhythmbox support

cheukyin699:


quinnuendo:
Isn't rhythmbox already available through MPRIS/DBUS?

cheukyin699:
Oh. I thought it was a plugin or something.

quinnuendo:
I haven't done a full test right now, but Rhythmbox does have MPRIS support, it might be through a plug-in, but I remember it working before.

Maybe a separate block with just Rhythmbox could make sense, for someone running just that program.

cheukyin699:
Yup. It does have mpris support, and thus can be controlled through `playerctl`.

kb100:
`playerctl` is an outside dependancy that rhythmbox users may not have already. It may make sense to include this PR anyway for that reason.

PR #105, opened 2017-08-13
battery: show time left to charge and check time validity

jolange:
- now time left to charge is also shown
- if time is unknown, acpi shows "rate information unavailable", leading
  to `(rate)` output, which is suppressed now

### DIFF
--- a/battery/battery
+++ b/battery/battery
@@ -43,17 +43,22 @@ else:
     # stands for unknown status of battery
     FA_QUESTION = "<span font='FontAwesome'>\uf128</span>"
 
-    timeleft = ""
-
-    if state == "Discharging":
-        time = commasplitstatus[-1].split()[0]
+    time = commasplitstatus[-1]
+    if time != "rate information unavailable":
+        time = time.split()[0]
         time = ":".join(time.split(":")[0:2])
         timeleft = " ({})".format(time)
+    else:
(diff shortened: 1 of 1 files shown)


kb100:
Good idea to include time until charged.

jolange:
> I just noticed this. Sometimes acpi gives the following confusing output
> ```Battery 0: Charging, 100%,  until charged```
> resulting in the blocklet writing a confusing `(until)`

That's a really a strange output... I think I'll simply submit a solution using a regex matching to avoid any other such unexpected things

kb100:
One easy fix would just be in the charging case (the `else` case) just check if percentleft is 100 and if so set timeleft back to empty

jolange:
But I guess when it's "charging" (and not "full"), the 100% is just a rounded value, so in principle there shoud be a time-left-to-charge value. As you said, that output is really confusing. Is that output permanent for you, when the battery is completely charged?

Anyway, I have a version using a regex match now. PR coming soon.

kb100:
You are right, I think it is rounded because eventually it does latch in to the "full" state. On the other hand, for whatever reason acpi does not provide a time-left-to-charge value. I guess a quick check of the format of timeleft being `\d+:\d+` is a more robust choice, otherwise null it.

PR #106, opened 2017-08-14
battery: explicitly match a time to avoid artifacts of confusing outputs

jolange:
(see #105)

### DIFF
--- a/battery/battery
+++ b/battery/battery
@@ -5,6 +5,7 @@
 #
 # A battery indicator blocklet script for i3blocks
 
+import re
 from subprocess import check_output
 
 status = check_output(['acpi'], universal_newlines=True)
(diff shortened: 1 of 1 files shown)


kb100:
Why not just

        if re.match("\d+:\d+", time):
            timeleft = " ({})".format(time)
        else:
            timeleft = ""

instead of overwriting `time` and then resetting it?

kb100:
Also it might be more readable to use a raw string since we have literal backslashes `re.match(r"\d+:\d+", time)` which I think is what we want.

jolange:
> Why not just (...)

That would also include all the rest of the output, which is for me something like `00:59:29 until charged`
> raw string

That's right, done!

PR #112, opened 2017-09-13
volume-pulseadio: add missing command in README

Q-Mart:
The config changes for media keys in the volume-pulseaudio is missing a command to restart i3blocks. This adds it


kb100:
Good catch, thanks and merged.

PR #113, opened 2017-09-14
another battery bar with configurable colors, screen locking, suspending, estimated times and a battery icon

vKnmnn:


vKnmnn:
i'm sorry, this seems to be getting messy. i'm still learning git.

kb100:
I think the code style in this blocklet is not up to the standards of this repository. To give constructive criticism, it looks like a lot of the block is copied from the existing batterbar code, which is fine, but the variable names, e.g. "squares" no longer represent what they did in that blocklet. I'm not trying to discourage you from contributing, but it seems like this blocklet is very similar to the existing battery blocklet https://github.com/vivien/i3blocks-contrib/tree/master/battery. You may have better luck by submitting small improvements to that blocklet (one new feature at a time) instead of creating a whole new blocklet. I suggest this because there are a lot of edge cases that we have slowly found and accounted for in that blocklet that are not accounted for in yours. If you are interested, the easiest thing to add to that blocklet would be configurable colors, then different icons for different battery levels. I think hibernating when low is a system task that should not be handled by a blocklet. This is usually configured in UPower.conf, see https://superuser.com/questions/1158459/does-laptop-linux-automatically-shut-down-hibernate-when-suspended-on-low-batt#1158465.

Also, when you create a pull request for a blocklet, make sure you only change one blocklet. In this pull request you have changed/added several. Pull requests should be as short and sweet a possible so they are easy to test and merge. Closing for now but if you are interested don't get discouraged, follow suggestions and I would definitely merge some of these changes.

kb100:
P.S. everyone learning git goes through this at some point. We are here to help (but research yourself before asking ;) )

vKnmnn:
alright then, how in the world do i open multiple PRs on your repo? it seems i can not work or commit on mine while the PR has not been dealt with. what is your suggested workflow?

kb100:
The key is to create a different git branch for each PR so they don't interfere.

PR #114, opened 2017-09-14
rofi-cal: handle long dates

vKnmnn:


kb100:
This commit contains merge garbage and will not run.

PR #116, opened 2017-09-30
time: Add new block

danstewart:
Adds a time block that changes timezone on click.

Two things:
1. Is there a better way to handle the default zones, or is expecting the user to edit the source alright?
2. Is there a good tool that can record a gif of the bar? Tried peek but it didn't show the bar in the gif.

Thanks


kb100:
1. Use the current system timezone, I'm sure perl has a function builtin for finding this. 2. No idea.

As far as what timezones to cycle through though, I think having the user edit the source is okay. Just make sure instructions in the source are clear on what it okay to change and what is not.

danstewart:
Thanks for the feedback.

I've added a default zone lookup and some basic comments around the two hashes that the user would be changing.

Thanks

kb100:
Looks good, just one more thing I think would be reasonable to add. Could you add an option to change the time string format? Perhaps use BLOCK_INSTANCE if you don't want to get into argument checking. It should be a simple change, see http://perldoc.perl.org/functions/localtime.html for what I'm talking about.

danstewart:
Now added. Thanks.

kb100:
Merged. I removed the extra time.pl. You may have seen this in other blocklets but this is strictly for backwards compatibility. New blocklets will be named without extensions.

PR #118, opened 2017-10-21
rofi-calendar: handle long dates + add spacing

vKnmnn:


Aster89:
@kb100, I suppose that posting here instead of opening an issue is a good idea when there is no issue. I hope I'm not wrong.

I've customized `rofi-calendar` in a way that it shows the next month upon right click (left and middle are kept on the current month). Would you be interested in a PR?

(My original idea was left=back1month, central=openclose, right=forth1month, so that one can move any number months back and forth, but I still don't have dynamic properties and I don't know how to keep track of the currently displayed month, so the most I could do is left=last, central=current, right=next, but I think people using it wouldn't like the change in leftbutton's action.)

vKnmnn:
open your own pull request. that way we can at least see your code. you can see that  this PR hasn't been touched in over a year

vKnmnn:
or PR it to me and i can update this one

PR #123, opened 2017-12-13
Remove local restriction from disk script

RX14:
Currently, the disk script fails to scan a disk if it's over the network. If a user has configured i3blocks to look at a specific disk, presumably he knows if it's local or not and can make a judgement as to whether he wants this behaviour. Consequently, we remove the -l option from df to allow remote disks to be monitored.

Perhaps this should be configurable, though.

### DIFF
--- a/disk/disk
+++ b/disk/disk
@@ -17,7 +17,7 @@
 DIR="${BLOCK_INSTANCE:-$HOME}"
 ALERT_LOW="${1:-10}" # color will turn red under this value (default: 10%)
 
-df -h -P -l "$DIR" | awk -v alert_low=$ALERT_LOW '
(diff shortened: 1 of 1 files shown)


kb100:
So I don't know why the original author included the `-l` flag, presumably because it makes the command faster/more efficient, or maybe just to suggest that pinging a network disk all the time is generally a bad idea, perhaps @vivien can comment. But I think I agree that a user should be able to do it anyway if they know. Even if a network disk hangs, i3blocks will not continually spawn new processes so at worst you will have out of date information. 

The only reason why I wouldn't simply remove the `-l` is because this is a default script, and in some situations a user's $HOME might be a network disk. For this reason, I'll just allow an override flag instead. Commit incoming shortly...

PR #125, opened 2018-01-03
kubernetes: add current context and namespace for kubernetes

dbenque:
new i3block to add current context and namespace for kubernetes


kb100:
I don't know about kubernetes but the script seems straightforward. This seems like a good candidate for customization options in the future (if you decide to improve upon the script). I'll merge as is for now, and in the future consider adding options to change the format/colors of the output instead of using hard coded symbols like ☸

PR #126, opened 2018-01-03
calendar: a date/time blocklet featuring a calendar popup created by yad

qw3ry:
I was not entirely content with the `rofi-calendar` script, so I decided to change it. Finally I decided the changes are worth a contribution.

I did not include a licence file (as requested in the contributing guidelines) because I do not care about a licence and the Repo is already licenced.

I did not test the blocklet on any machine but my own and I am not sure if some things break there. Still, there should be easy fixes for that.


kb100:
When I click on the blocklet it doesn't come up as a floating window. Is there some way to make the window floating? This seems to be the intention of how the block should work.

qw3ry:
I am not sure how the window hinting in i3 works. It seems to work if I add the `--splash` option, but the manpage says it would set `_NET_WM_WINDOW_TYPE_SPLASH` which doesn't seem appropriate.

I [filed a bug](https://sourceforge.net/p/yad-dialog/tickets/301/) at yad.

I don't know if it is better to set the `--splash` option - which works for me, but might break if i3 changes the way it handles splash screens or if someone uses another wm (is that possible?) - or instruct the user in the readme to add `for_window [class="Yad"] floating enable` to the i3 config (that is indeed how I am using it now, but I feel it is not a good idea to require additional configuration if its not neccessary)

What do you think?

kb100:
I would just add a note in the README to add your `floating enable` line to your i3 config, and then when yad fixes this bug just remove the note from the README. If the yad people mark it as intended behavior then just use the splash flag.

qw3ry:
@kb100 done.

PR #127, opened 2018-01-03
battery3: yet another battery blocklet

qw3ry:
This blocklet displays the battery status enhanced with fontawesome icons.
It is basically a copy of the default battery script.

I was not sure what to do with the copyright/licence notice in the header. I removed it for now because this repo is licenced under gpl itself. Feel free to complain about this though.


kb100:
Closing because this doesn't provide anything new (as far as I can see) beyond what the battery2 script does.

PR #129, opened 2018-01-05
volume: don't check for pulse via lsmod but command

jolange:
- check if command `pulseaudio` exists
- check if pulse is running

Fixes vivien/i3blocks#236
Fixes vivien/i3blocks#201

Replaces original PR vivien/i3blocks#237

### DIFF
--- a/volume/README.md
+++ b/volume/README.md
@@ -36,3 +36,7 @@ instance=Master
 interval=once
 signal=10
 ```
+
+The script takes two optional parameters: step size and mixer selection, e.g. `$SCRIPT_DIR/volume "5%" default`.
+By default the script tries to make a reasonable choice for the mixer. If it does not work for your setup (e.g. for (un)muting), try to figure out the parameter using `amixer info`.
+
--- a/volume/volume
+++ b/volume/volume
@@ -18,12 +18,17 @@
 #------------------------------------------------------------------------
 
 # The second parameter overrides the mixer selection
-# For PulseAudio users, use "pulse"
+# For PulseAudio users, eventually use "pulse"
 # For Jack/Jack2 users, use "jackplug"
 # For ALSA users, you may use "default" for your primary card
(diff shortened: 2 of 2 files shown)


kb100:
Hmm... this blocklet crashes for me, both in the current version and the new version. I'm using pulseaudio, but there is no "pulse" mixer. `amixer scontrols` returns

    Simple mixer control 'Master',0
    Simple mixer control 'Capture',0

The blocklet crashes with

    ALSA lib control.c:1373:(snd_ctl_open_noupdate) Invalid CTL pulse
    amixer: Mixer attach pulse error: No such file or directory
    Specify a scontrol identifier: 'name',index
    Specify a scontrol identifier: 'name',index

`alsamixer` shows

    Card: PulseAudio
    Chip: PulseAudio

I'm not sure if this is due to an update in how pulse/alsa interact, or if it is a debian unstable bug, or if I setup my sound in a nonstandard way (if I did, it was years ago, I don't remember ever setting it up). In any case, calling the script as `./volume "5%" default` fixes the issue for me...

At the very least we should include instructions for what to do if you get this error, if not circumvent it somehow from the script. What are your thoughts?

jolange:
Indeed I have the same problem on a tumbleweed machine with pulse... It's a mess :-D

I will try to figure out a way to catch all cases and additionally add the manual parameter settings to the README.

kb100:
https://github.com/vivien/i3blocks-contrib/blob/master/volume-pulseaudio/volume-pulseaudio Also suffers from this bug.

We need to figure out a reliable way to determine what the mixer and scontrol need to be.

jolange:
Ok, I've got a version that works for the cases I was able to test. @kb100 could you try this version?

_[I will squash and rebase this once we are done]_

kb100:
Works for me.

jolange:
Great. Everything is squashed and rebased.

PR #130, opened 2018-01-05
example config: (cpu_usage) fix block min_width

jolange:
The default block changed from `CPU: 100.00%` to `CPU 100.00%` in vivien/i3blocks@1209af4, i.e. the colon is not used in the label.

Replaces original vivien/i3blocks#222

### DIFF
--- a/config.example
+++ b/config.example
@@ -93,7 +93,7 @@ interval=5
 [cpu_usage]
 label=CPU
 interval=10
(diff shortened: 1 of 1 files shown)


kb100:
Thanks, merged. In the future please start commits with the name of the blocklet, e.g. cpu_usage: blah blah blah.

PR #131, opened 2018-01-05
Fix bug with incorrect regex

merspieler:
Signed-off-by: merspieler <merspieler@users.noreply.github.com>

Fixed bug with wrong regex causing an error message that no GPU was found.


kb100:
Merged, thanks. In the future please begin commit messages with the name of the blocklet, in this case `gpu-load: Fix bug with incorrect regex`.

merspieler:
ok, i'll remember that

PR #133, opened 2018-01-11
keyboard-layout: add keyboard-layout, shows currently active keyboard layout

merspieler:
Advantage, compared to the existing one: When there are multiple layouts, the other script prints `us, es` no matter which one is active.

Disadvantage: Additional dependency.

Signed-off-by: Fabian Lucke <flucke@data-modul.com>


merspieler:
Closed to apply lastminute bugfix.

PR #134, opened 2018-01-11
keyboard-layout: add keyboard-layout, shows currently active keyboard layout

merspieler:
Advantage, compared to the existing one: When there are multiple layouts, the other script prints us, es no matter which one is active.

Disadvantage: Additional dependency.

Signed-off-by: merspieler <merspieler@users.noreply.github.com>


kb100:
What does this blocklet provide beyond the existing layout script https://github.com/vivien/i3blocks-contrib/tree/master/kbdd_layout?

merspieler:
It keeps the default keyboard switching functionality. kbdd is (at least i understood it that way) window based, the standart system tool is global. So if a global layout is prefered, this script can be used, who wish, to have window based layout can use the kbdd script.

After all, you're the maintainer and you can decide, if it makes sense, to include this script. I'll keep using this no matter, if it's in this repo or not. I just want to help others, who are looking for a possible solution.

kb100:
I see. I have been struggling to decide on how to deal with keyboard layouts since this repo started. I think that it is reasonable for someone to prefer a global config over a per-window config, so this PR will probably be merged. Please give me some time to test the blocklet since I don't normally use more than one keyboard layout.

kb100:
I have been wondering though, if you prefer a global keymap instead of a windowed one, would it not be easier to just use the basic key_layout block and put, e.g.

    bindsym Ctrl+Alt+1 exec setxkbmap -layout us
    bindsym Ctrl+Alt+2 exec setxkbmap -layout ru

in your i3 config (assuming you want e.g. English and Russian)? This wouldn't require any other dependencies either. Please let me know your opinion.

merspieler:
It's an option as well, it just doesn't do the cycling (which i prever since i habe a bunch of layouts depending on language and purpose)

merspieler:
@kb100 will it be merged or can i close the pr?

Issue #132, opened 2018-01-08
Update all blocklets to be configured via injected config properties (next branch)


kb100:
In the upcoming release of i3blocks, arbitrary properties will be able to be injected into a blocklet's environment. The readability of the i3blocks configuration file will be substantially improved by making all block configuration done through these injected environmental variables.

By way of example, the volume-pulseaudio blocklet previously had an example config:

 
```
[volume-pulseaudio]
command=$SCRIPT_DIR/volume-pulseaudio -a
interval=once
signal=1
```

This should be replaced by something like

```
[volume-pulseaudio]
command=$SCRIPT_DIR/volume-pulseaudio
interval=once
signal=1
USE_ALSA_NAME=1
```

More generally, the README should include something like:

```
[volume-pulseaudio]
command=$SCRIPT_DIR/volume-pulseaudio
interval=once
signal=1
# exposed format variables: %s symbol, %v volume, %i index, %n name
#FORMAT_LONG=%s %v\% [%i:%n]
#FORMAT_SHORT=%s %v\% [%i]
#SUBSCRIBE=0
USE_ALSA_NAME=1
#USE_DEVICE_NAME=0
#AUDIO_HIGH_SYMBOL='  '
#AUDIO_MED_SYMBOL='  '
#AUDIO_LOW_SYMBOL='  '
#AUDIO_MUTED_SYMBOL='  '
#AUDIO_MED_THRESHOLD=50
#AUDIO_LOW_THRESHOLD=0
#COLOR_NON_MUTED=#ffffff
#COLOR_MUTED=#a0a0a0
```

For each blocklet, we must:

1. Update the script itself to use environmental variables for all possible configuration. Command line argument code should not be deleted if possible in order to avoid breaking existing configurations.
2. Update the README to include an example config which shows all configurable parameters. Parameters that are not used in the example config should still be included, but commented out with their default value.
3. Update all example configs with the new syntax.

Note that while we are at it, now is a great time to implement format string options `FORMAT_LONG` and `FORMAT_SHORT` as above, or allowing different formats like `FORMAT_UP` and `FORMAT_DOWN`.
This can cut down on a lot of format specific configuration options. For instance, in the volume-pulseaudio block above, we could have `FORMAT_HIGH, FORMAT_MED, FORMAT_LOW, FORMAT_MUTED` and just let the user do exactly what they want.

All work done towards this end should be on the next branch, not the master branch.
If you want to help, submit pull requests against the next branch.

[...]

PR #135, opened 2018-01-13
calendar: added options for format, width and height

jolange:
@qw3ry these changes are completely backwards compatible, but maybe you still want to check them.

### DIFF
--- a/calendar/README.md
+++ b/calendar/README.md
@@ -32,8 +32,13 @@ You may skip this step if you want to open the calendar popup as a normal tile i
 
 # Configuration
 
-You may want to
+Parameters:
 
-* edit the date/time format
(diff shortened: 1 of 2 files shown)


kb100:
I like the idea of having these as configurable, but for future blocklets I would like to have configuration done through inejcted environmental variables. See https://github.com/vivien/i3blocks-contrib/issues/132. Could you update this commit and PR against the 'next' branch?

jolange:
Sure, I will do that. The question is when we expect the corresponding i3blocks release and if we still want to have this one merged for the current version.

qw3ry:
@jolange thanks for the heads up, thats definitely a good idea 👍

PR #137, opened 2018-01-17
volume: pulse-detection: explicitly try the "pulse" device

jolange:
I noticed on one machine (where pulse is the right option) that it's not the default output of `amixer info` and you need to explicitly specify `-D pulse`. When this command succeeds, there is no more need to check the output.

### DIFF
--- a/volume/volume
+++ b/volume/volume
@@ -23,9 +23,9 @@
 # For ALSA users, you may use "default" for your primary card
 # or you may use hw:# where # is the number of the card desired
 MIXER="default"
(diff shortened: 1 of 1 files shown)


kb100:
Thanks for this. Merged and also added the same changes to volume-pulseaudio. I bet vivien needs testers for his next branch and I also need a lot of help converting blocklets for my next branch. If you want to help, you could probably start with updating the volume blocklet for the next branch. See https://github.com/vivien/i3blocks-contrib/issues/132 if interested.

jolange:
Yeah, it's on my TODO list to set up the new system!

PR #139, opened 2018-02-01
aur-update: new blocklet to check for AUR updates

jolange:
List updates from the Arch User repository.

(If this should be merged, I'll prepare a corresponding PR for the next branch)


kb100:
I'd actually like to freeze the master branch for now except for bugfixes. Should have mentioned this to you earlier, so all new PRs should be against the next branch and use injected properties instead of command line switches. Also, I don't use arch so could you explain the difference of this blocklet and the arch-update blocklet? Would it be possible to merge any new features instead of creating a new block?

jolange:
Here's some context: The normal pre-compiled packages for arch are installed via the _pacman_ package manager, which downloads from specific repositories. This is what the existing arch-update blocklet checks.
In addition, users can use so-called _PKGBUILDs_ to create recipes for building programs from source and convert them into a package file, which in turn can then be installed with _pacman_. The AUR is a place to collect user-provided PKGBUILDs that might be of interest for other users. The "normal" workflow is to download this, run a command that executes the recipe (i.e. compiles and creates a binary package) and then install the package. This blocklet checks the installed packages for ones that come from the AUR (i.e. are not installed from the normal package repos) and then compares to the latest version in the AUR.

There are also programs to automate AUR package installation that provide pacman-like usage. These are not official distro tools, though, and not used by everyone. The existing arch-update blocklet has an AUR option that uses one such tool (_yaourt_) already.

I personally strongly prefer to keep the normal packages and the AUR ones completely apart, because available updates have rather different implications. For the normal ones, I just run a simple pacman "update my system". For the AUR ones, I have to become active on my own and update them *singly* (or choose to ignore them and keep my version).

If that makes sense to you, I'll prepare a PR against next.

kb100:
I see the benefit now. It sounds like arch users will be familiar with this. Sounds good to me. Go ahead and prepare the PR against next.

PR #141, opened 2018-02-16
redefine critical state

kenardes:
what is call "critical" should be where the internet speed low to zero.

### DIFF
--- a/bandwidth2/bandwidth2.c
+++ b/bandwidth2/bandwidth2.c
@@ -72,9 +72,9 @@ void get_values(char *const iface, time_t * const s, ulli * const received, ulli
 
 void display(int const unit, double b, int const warning, int const critical)
 {
-  if (critical != 0 && b > critical) {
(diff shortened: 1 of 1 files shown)


kb100:
I believe you misunderstand the purpose of the critical and warning values. It does not make sense to warn the user when their internet usage is low because internet usage is very often near zero. The warnings are supposed to be for when your usage is too high, with the use case in mind being a person with limited bandwidth that wants to be aware whenever they are using a lot of bandwidth.

kenardes:
in my case, i use my android's tethering to access internet, since the signal not that strong, i  need to make sure the connection and internet speed to always good, so for me cricital = poor signal / low speed.

kb100:
I believe that I understand your concern, but the bandwidth2 block measures neither signal strength nor connection speed. Its output is the rate at which bytes are are currently being sent/received. It does not perform any kind of speed test.

PR #153, opened 2018-04-06
Added shutdown_menu error messages

stevenengler:
If an error occurs (such as when hibernating), a notification using 'notify-send' will alert the user.


kb100:
Unfortunately this solution is not workable. Directly calling ${menu[${selection}]} freezes the bar, so this is not acceptable. On the other hand, i3-msg exec -q does not capture the exit code of a command, so it cannot be used to test either. If a workaround is possible, it must use i3-msg exec -q to spawn a shell that runs the program and then notify-send if the program fails.

Issue #21, opened 2015-12-16
Is it possible to use the openvpn block with networkmanager-openvpn?


sultanofrum:
I want to use networkmanager to manage my OpenVPN connection, as it's the only way I can get it to reliably restart after suspend. However, with this setup, the OpenVPN script in this repository just says 'down', even though vpn is working.

I can change the dev name of the vpn connection to 'tun0' instead of 'tun' in /etc/NetworkManager/system-connections/_vpn_id_ , but I don't know how to make it write a pid file.

Is there a way to make the [openvpn] block work with my setup?

Or is it easy enough to make a new block for this? i3status for example supports networkmanager-openvpn setups with this sort of configuration:

```
path_exists VPN {
        # path exists when a VPN tunnel launched by nmcli/nm-applet is active
        path = "/proc/sys/net/ipv4/conf/tun0"
}  
```

kb100:
If all path_exists is doing is checking whether the file exists, that can be emulated in i3blocks with

    [vpn]
    command=[ -e $BLOCK_INSTANCE ] && (echo up; echo up; echo #00FF00) || (echo down; echo down; echo #FF0000)
    interval=10
    instance=/proc/sys/net/ipv4/conf/tun0

sultanofrum:
Thank you very much, it works and is a lot more elegant than what I came up with!

I understand that just checking the path is much less sophisticated than what the openvpn blocklet in this package does, but that is all I can really manage at this state, having no knowledge in perl and almost none in bash or linux.

kb100:
Perhaps @deimosfr (as far as I can tell, the original author of the openvpn script) would be interested in modifying his script to be compatible with networkmanager-openvpn? I personally have never used a vpn and thus am not well equipped to rewrite it myself.

deimosfr:
Hi, yes I think I'm going to take a look about it as I'm using network manager too. I'll keep you updated

sultanofrum:
Thank you very much, I am looking forward to it!

deimosfr:
Hi,

I'm currently testing a quick trick for OpenVPN. Add this in a simple bash script file:
```
nmcli -t -f type,state,connection d | egrep '^tun|tap' | grep connected | awk -F':' '{ print "VPN: " $3 }' || echo 'VPN: down'
```
This requires nmcli of course. It doesn't support multiple VPN connections as NetworkManager doesn't too. Please let me know if it works for you too

sultanofrum:
[...]

PR #155, opened 2018-04-28
nm-vpn: add nm-vpn, a script to show vpn status via nmcli

The-King-of-Toasters:
I've been experimenting with i3blocks and decided to create a AWK program to solve #21.

It grabs output from `nmcli` and prints the VPN name/status in green.


The-King-of-Toasters:
Hmm, it seems I've done something very wrong here. I'll make a fresh pull request.

PR #156, opened 2018-04-30
calendar: adds support for calendar popup position top/bottom

jose-lopes:
I use i3bar with position top, so I needed to change this block to support popup position to be top or bottom.
I added a parameter to configure the position of  the popup.


jolange:
The option makes sense! But the master branch is currently frozen. Can you submit a PR against the next branch?

calendar/calendar
```diff
@@ -22,11 +24,15 @@ while getopts ":f:W:H:" opt; do
 done
 
 case "$BLOCK_BUTTON" in
-  1|2|3)
```
jolange: Please don't introduce diff noise.

jose-lopes:
@jolange Ok i will do it.

PR #159, opened 2018-05-26
battery2: add support for colors of fontawesome icons

medik:
Hi! This is my first contribution to i3blocks!

## What is new?
i3blocks-diff
*(Top: Before; Below: Now)*
I've modified battery2 with support for colored font awesome icons and are searching for feedback. Note that the code I've written might be considered as "messy", but can be easily fixed if requested.

## Are there better ways to encode the colors?
i3blocks-code
As seen above, I have encoded the colors into global variables inside the code and the question is whether there are better ways to edit the colors instead of to hardcode the colors. I've read the [documentation](https://github.com/vivien/i3blocks/wiki/Writing-Your-Own-Blocklet) where it says that the script can access the user configuration input via the environment variables `BLOCK_NAME`, `BLOCK_INSTANCE` and the other three is to access mouse input (which are irrelevant to this case). The documentation doesn't say it explicitly under the Environment Variables subtitle, but I guess that the `instance=...` in each block is where you enter the value which are sent to `BLOCK_INSTANCE`? Furthermore, what is the best practise for sending more than one value? Here are two suggestion I have:
```
{'lightning_color': 'blue', 'plug_color': 'pink'}
blue,pink
```

The first one is *ugly*, but more readable than the last one. It is ugly because it doesn't look consistent with the previous configurations in i3blocks.conf. And is readable because I know which one is which without having to look in the documentation or the source code every time I want to change something.

What I want to say here is that we need better ways to send multiple values in i3blocks. Please.

Thank you for taking your time to reading about me ranting :)

### DIFF
--- a/battery2/battery2
+++ b/battery2/battery2
@@ -5,14 +5,30 @@
 #
 # A battery indicator blocklet script for i3blocks
 
+# Copyright (C) 2018 Olof Sjödin <me@olofsjodin.se>
+# Edits are licensed under the GPL version 3
+#
+# Added support for colored icons.
+
+
 import re
 from subprocess import check_output
 
 status = check_output(['acpi'], universal_newlines=True)
 
+# Colors
+BATTERY_NO_FOUND_COLOR = "red"
+LIGHTNING_COLOR = "yellow"
+PLUG_COLOR = "cyan"
+BATTERY_COLOR = "cyan"
+UNKNOWN_STATUS_COLOR = "cyan"
+
+def add_color(color, body):
+    return "<span color='" + color + "'>" + body + "</span>"
+
 if not status:
     # stands for no battery found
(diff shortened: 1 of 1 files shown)


kb100:
Hi @medik, thanks for this idea. Currently the [master] branch is frozen, so all PRs should be against the [next] branch. The [next] branch's goal is well-suited for what you want to do in this commit. The [next] branch of i3blocks allows for arbitrary injected environmental variables, and the [next] branch of i3blocks-contrib has updated all the scripts to use these injected variables. See https://github.com/vivien/i3blocks-contrib/blob/next/monitor_manager/monitor_manager for an example of how to use these in python, and see https://github.com/vivien/i3blocks-contrib/issues/132 for more info about what the [next] branch is trying to achieve.

medik:
@kb100 Thank you for your comment. I'll get to it ASAP. 

I will close this and open a new PR when it is ready.

/Olof

PR #161, opened 2018-06-03
arch-update: add support for yay AUR helper

Q-Mart:
Add support for yay AUR helper by modifying the -a argument. This also creates a way for support for other AUR helpers

### DIFF
--- a/arch-update/arch-update.py
+++ b/arch-update/arch-update.py
@@ -9,6 +9,8 @@ from subprocess import check_output
 import argparse
 import re
 
+class AURHelperNotSupportedError(Exception):
+    pass
 
 def create_argparse():
     parser = argparse.ArgumentParser(description='Check for pacman updates')
@@ -27,7 +29,6 @@ def create_argparse():
     parser.add_argument(
         '-a',
         '--aur',
-        default='yaourt',
         help='Include AUR packages and specify which AUR helper you use. '
         'Attn: Yaourt and yay are supported and must be installed'
     )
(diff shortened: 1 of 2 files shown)


arch-update/arch-update.py
```diff
@@ -27,8 +27,9 @@ def create_argparse():
     parser.add_argument(
         '-a',
         '--aur',
-        action='store_true',
-        help='Include AUR packages. Attn: Yaourt must be installed'
+        default='yaourt',
```
jolange: This means that by default AUR updates are checked using yaourt, even if `-a` is not specified.

jolange:
In principle a good  idea to support different tools, but the changes should be backwards-compatible. Please note that PRs should be submitted against the "next" branch.

arch-update/arch-update.py
```diff
@@ -59,17 +60,29 @@ def get_updates():
...
+
+    elif aur_helper == 'yay':
+        try:
+            output = check_output(['yay', '-Qua']).decode('utf-8')
+        except subprocess.CalledProcessError as exc:
+            # yay also exits with 1 and no output if no updates are available
+            if not (exc.returncode == 1 and not exc.output):
+                raise exc
+        if not output:
+            return []
```
jpleau: It is perhaps a wise idea to DRY the code a bit here.

The only difference between "yaourt" and "yay" is the executable (and maybe the arguments).

The entire logic to prepare the command, run it, catch exceptions shouldn't be duplicated (imagine if we had 10 AUR helpers)

Q-Mart:
Thanks for your comments everyone, I've updated the code according to them and created a new PR that merges into next (#163)

PR #167, opened 2018-07-17
Add missing dependency in arch-update script

joelwurtz:


jolange:
This was actually included in the next branch already: 7c1ec53

joelwurtz:
Didn't know there was a next branch, you can revert this then

jolange:
It's not a problem, no need to revert ;-) Will just need to be merged at some point

kb100:
I actually merged this, forgetting that I had already merged a similar PR into the next branch and fixed it up in the next branch. I think this was an important one to merge into the master branch immediately though because if a user doesn't have the right deps then the whole block is unusable.

PR #168, opened 2018-08-12
[RFR] Docker block

digitronik:
While working with my task I like to use this docker block. Maybe other also interested in it.
Simple but very useful. while working with docker showing the `number of running containers` and `latest container IP address`.

Other things are possible but I don't like to make this bulky.


kb100:
Short and sweet block. Thanks for the contribution. I've merged it into the [next] branch (master is frozen, so in the future please submit blocklets against the next branch).

digitronik:
@kb100: awesome :+1:

Issue #145, opened 2018-03-19
OpenVPN block not detecting VPN even when it's working.


da-edra:
### Expected behavior
See the current status of my VPN on i3blocks.

### Actual behavior
OpenVPN is up and running, it shows down on i3blocks.

### i3blocks config relevant to blocklet(s)

```
[openvpn]
command=~/.i3/i3scripts/openvpn
instance=tun0
interval=1
```

### Output of blocklet(s) when run from command line
VPN: down
VPN: down

### Output of any relevant other commands that might help diagnostics
**systemctl status openvpn-client@japan.service**

``` ● openvpn-client@japan.service - OpenVPN tunnel for japan
   Loaded: loaded (/usr/lib/systemd/system/openvpn-client@.service; enabled; vendor preset: disabled)
   Active: active (running) since Mon 2018-03-19 01:14:54 CST; 3min 42s ago
     Docs: man:openvpn(8)
           https://community.openvpn.net/openvpn/wiki/Openvpn24ManPage
           https://community.openvpn.net/openvpn/wiki/HOWTO
 Main PID: 31332 (openvpn)
   Status: "Initialization Sequence Completed"
    Tasks: 1 (limit: 4915)
   Memory: 3.8M
   CGroup: /system.slice/system-openvpn\x2dclient.slice/openvpn-client@japan.service
           └─31332 /usr/bin/openvpn --suppress-timestamps --nobind --config japan.conf --writepid /var/run/openvpn-client/japan.pid

mar 19 01:14:57 Oblivion openvpn[31332]: /etc/openvpn/scripts/update-systemd-resolved tun0 1500 1585 10.8.0.7 255.255.0.0 init
mar 19 01:14:57 Oblivion openvpn[31332]: <14>Mar 19 01:14:57 update-systemd-resolved: Link 'tun0' coming up
mar 19 01:14:57 Oblivion update-systemd-resolved[31888]: Adding IPv4 DNS Server 10.8.0.1
mar 19 01:14:57 Oblivion openvpn[31332]: <14>Mar 19 01:14:57 update-systemd-resolved: Adding IPv4 DNS Server 10.8.0.1
mar 19 01:14:57 Oblivion update-systemd-resolved[31889]: SetLinkDNS(6 1 2 4 10 8 0 1)
mar 19 01:14:57 Oblivion openvpn[31332]: <14>Mar 19 01:14:57 update-systemd-resolved: SetLinkDNS(6 1 2 4 10 8 0 1)
[...]

PR #170, opened 2018-08-31
openvpn2: simpler version of openvpn using nmcli

deimosfr:
According to issue #145, I'm proposing a simpler solution based on nmcli

### DIFF
--- a/openvpn2/README.md
+++ b/openvpn2/README.md
@@ -1,0 +1,16 @@
+# openvpn
+
+Support OpenVPN (require nmcli binary)
+
+# Usage
+
+Add the following to your i3blocks config:
+
+``` ini
+[openvpn]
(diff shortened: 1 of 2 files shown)


kb100:
Hi, thanks for creating this. I think with a few changes this would be a great addition to the repo. I don't use openvpn or network manager, so would you please create a clear list of dependencies for this block. For instance, presumably openvpn should be a dependency, and as far as nmcli goes, what package does this usually come in? apt search nmcli turned up nothing.

Finally, the master branch is currently frozen, so please close this pull request and submit a new one, making sure you have tested the block against the [next] branch. See https://github.com/vivien/i3blocks-contrib/issues/132 for details on this branch.

da-edra:
Closing because we've already implemented this functionality. Thanks!!

PR #172, opened 2018-09-19
Handle changes of default interface in bandwidth3.

nolanl:
Adds an "autoreload" or '-a' option that enables bandwidth3 to
automatically handle changes of the default iterface/default route.


nolanl:
I have switched to bandwidth2, and have a better PR for that block.

PR #180, opened 2018-10-28
Exit on missing default route

bl1nk:
When there's no default route the scripts should exit early.  I noticed the `iface` script erroring when my network was down.

From what I can tell `iface` and `bandwidth` are affected. `bandwidth3` also uses `ip route | awk` to get the interface name but deals with it later so no change was required there.


kb100:
Thanks for this, merged! I changed the commit message. Next time include the names of modified blocks in the commit message.

PR #181, opened 2018-10-29
memory: no newline after foreground color

jolange:
because no background color is printed


jolange:
Yes, this produces a background color for the blocklet:

2018-10-29-152838_1920x1200_scrot

kb100:
Thanks for this! Merged.

PR #189, opened 2018-12-05
iface: change iface script, want to only add first word to clipboard …

zuntik:
…in case the connection is tunneled
Example: activate a VPN connection and will see: "<ip1> peer <ip2>"

### DIFF
--- a/iface/iface
+++ b/iface/iface
@@ -61,7 +61,7 @@ fi
 IPADDR=$(ip addr show $IF | perl -n -e "/$AF ([^\/]+).* scope global/ && print \$1 and exit")
 
 case $BLOCK_BUTTON in
-  3) echo -n "$IPADDR" | xclip -q -se c ;;
+  3) echo -n "$IPADDR" | sed -e 's/\s.*$//' | xclip -q -se c ;;
 esac
 
 #------------------------------------------------------------------------


kb100:
Would you please post the output of `ip addr show $IF` where `$IF` is the interface you are using. This fix does not solve the root of the problem which is that for some reason the `$IPADDR` variable, which should be an ip address, contains extra garbage in your case.

zuntik:
the output you want
`ip addr show $IF`
```
4: tun0: <POINTOPOINT,MULTICAST,NOARP,UP,LOWER_UP> mtu 1500 qdisc fq_codel state UNKNOWN group default qlen 100
    link/none
    inet xx.xx.xx.xx peer xx.xx.xx.xx/32 brd xx.xx.xx.xx scope global noprefixroute tun0
       valid_lft forever preferred_lft forever
    inet6 2001:xx:xx:xx::xx peer 2001:xx:xx:xx::1/64 scope global noprefixroute
       valid_lft forever preferred_lft forever

```

kb100:
@zuntik https://github.com/vivien/i3blocks-contrib/commit/3af068b9c24aecdf541418123e812a4e3489b714 should fix your issue. Would you please test it and let me know if it indeed fixes the problem?

PR #190, opened 2018-12-11
rofi-arch-update: rofi list of available updates upon click on blocklet

Aster89:


Aster89:
I do not use the `git` version of `i3blocks` so I can't check if everything works with the dynamic properties. Hope it does.

Issue #191, opened 2018-12-17
cpu_usage: cannot set to 0 decimal places


jaimelive:
cpu_usage does not support setting variables, such as DECIMALS to 0

```
[cpu_usage]
label=CPU: 
interval=10
DECIMALS=0
```

This is a consequence of improperly using the OR operator to set default values:
```
# default values
my $t_warn = $ENV{T_WARN} || 50;
my $t_crit = $ENV{T_CRIT} || 80;
my $cpu_usage = -1;
my $decimals = $ENV{DECIMALS} || 2;
my $label = $ENV{LABEL} || "";

```
0 || 2 = 2

PR #192, opened 2018-12-17
cpu_usage: Fixes: #191 - properly set defaults

jaimelive:
Changed the OR operation to the // default operator. This way, values can be set to 0.


kb100:
Thanks for this!

PR #193, opened 2018-12-23
Add multiple interfaces support to bandwidth2.

nolanl:
This change allows you to specify a list of interfaces to watch,
separated by commas.  If you don't specify any interfaces, the
existing behavior is preserved, and all interfaces (except lo) are
summed.

### DIFF
--- a/bandwidth2/bandwidth2.c
+++ b/bandwidth2/bandwidth2.c
@@ -1,3 +1,4 @@
+#define _DEFAULT_SOURCE
 #include <stdio.h>
 #include <stdlib.h>
 #include <time.h>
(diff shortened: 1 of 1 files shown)


nolanl:
Edited commit message to match CONTRIBUTING.md requirements.

nolanl:
Thanks for the review!

> What is the purpose of the #define _DEFAULT_SOURCE?

strdup and strsep are POSIX and BSD (respectively).  The Makefile specifies -std=c11, so only ANSI C functions are allowed.  Given that the code is already very Linux-specific (thanks to parsing /proc/net/dev), adding a POSIX/BSD dependency seemed reasonable.

An alternative that might be less ugly is to specify -std=gnu11, which enables POSIX and BSD functions (and GNU functions, but the code doesn't use any of those).

> Also I have noticed that you have opted to use heap memory (calloc and strdup). Is this necessary?

Not strictly necessary.  I did it because I didn't want to artificially limit the number of interfaces.  I could simply pick a max number of interfaces and statically allocate the array.

> Case in point, neither your strdup or calloc'd memory is freed, so if this program runs long enough it will crash your GUI.

This is not actually the case.  The memory is allocated at startup, and used for the life of the program.  This is somewhat hidden by the strsep; the strings that make up the array of interface names are actually fragments of the strdup'd string.  strsep returns pointers into the middle of the source string, and writes '\0' over the token separator characters, so it doesn't need to allocate internally for each token.

The logically correct place to free the memory is after the while loop, right before program exit, but as this program never exits, there seemed to be no point; anything after the while(1) loop is dead code.

nolanl:
Looks great, thanks for the review, the fix and merge!

PR #194, opened 2018-12-25
volume-pulseaudio: adds tracking of running device instead of the default one

linosgian:
## Description

When switching devices volume-pulseaudio does not switch to the newly activated device.
As a result, it's stuck with the default device's volume, mute status etc.

That happens because volume-pulseaudio keeps track of the default device instead of the device with that has the *RUNNING* state, according to *pactl*'s output.

If no device is being used (all devices are on *SUSPENDED* state), I fall back to the default one since pactl/pacmd do not give information about which of the suspended devices is the currently used one.
Once any of them goes to *RUNNING* state though, will be picked up by the script. 

## Proposed fix
This pull request fixes this issue by grepping for the device that's on *RUNNING* state and falling back to the previous solution when all devices are suspended.
I have tested this with my bluetooth headset and several other output devices.

### DIFF
--- a/volume-pulseaudio/volume-pulseaudio
+++ b/volume-pulseaudio/volume-pulseaudio
@@ -118,9 +118,11 @@ function print_format {
 }
 
 function print_block {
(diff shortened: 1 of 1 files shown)


kb100:
Thanks for this contribution!

PR #195, opened 2018-12-27
nm-vpn: add nm-vpn, shows vpn status via nmcli

The-King-of-Toasters:
This is a redo of #155, as it's been so long since that commit was made that it would screw up the project history. I've simplified it and removed the bashisms. I also changed the licence to MIT, but I'm willing to change it back if needed.

This is a simple AWK program that reads input from `nmcli` and prints the currently connected VPN in green.

Closes: #21


The-King-of-Toasters:
goddammit not again, lemme do this one more time.

PR #198, opened 2019-01-11
Add new (C) cpu_usage2 blocklet.

nolanl:
Pretty much the same as cpu_usage, but written in C, and interval=persist.

Useful for folks who want to have per-second CPU utilization stats without burning a ton of CPU.


kb100:
Very nice and clean block, thanks for this. Out of curiosity, was the other cpu block actually burning a ton of CPU? On all machines I have tested it appears essentially idle.

nolanl:
Thanks for reviewing and merging!

Calling it "a ton of CPU" could be unfairly slandering the perl implementation, but it did show up as a couple of % when I set it to interval=1.  Perhaps I need a new laptop.  =)

PR #200, opened 2019-01-20
luminosity

m-242:
A blocklet using intel_backlight to show a percentage of the screen brightness.


kb100:
Would you mind if the name is changed to intel_backlight?

kb100:
Or backlight_intel

m-242:
No, not at all

PR #203, opened 2019-01-21
wifi: echo red "down" when interface exists but inactive

Maverobot:
The current `wifi` script does not show "wifi down" when the wifi device i.e. interface exists but is not active, which is a common situation on a laptop. Besides, this pull request makes the behavior of `wifi` script consistent with `iface` script.


kb100:
Good catch, nice clean commit.

kb100:
In that case, welcome to the open source community!

PR #204, opened 2019-01-24
Remove `label` from list of valid properties

hseg:
After vivien/i3blocks@61646f2, this is no longer valid

### DIFF
--- a/config.example
+++ b/config.example
@@ -11,7 +11,6 @@
 # full_text
 # instance
 # interval
-# label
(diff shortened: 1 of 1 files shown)


kb100:
Label support was actually brought back because this change affected so many users. It will eventually be phased out again though.
https://github.com/vivien/i3blocks/commit/37f23805ff886639163fbef8aedba71c8071eff8

PR #207, opened 2019-02-04
bandwidth2: fix padding and add SI units

nolanl:
Two commits, 2 fixes, 1 feature.

One fix is to make space padding work when a LABEL is specified.  The feature is to add support for SI units.  And the other fix is to make the amount of padding correct for non-SI units.

More details are in the individual commits' messages.

### DIFF
--- a/bandwidth2/README.md
+++ b/bandwidth2/README.md
@@ -9,6 +9,7 @@ It comes with some other features though:
 * Automatically estimate what unit (K,M,G,T) to use depending on the value. You can still choose between bits and bytes.
 * By default sum all the network interfaces (except lo) instead of only default route interface.
 * Warning and critical colors as an option.
+* Choice for SI units or binary units.
 
 ## Build
 
@@ -25,11 +26,12 @@ markup=pango
 LABEL=
 #INTERFACE=(checks all interfaces)
 #USE_BITS=0
-#USE_BYES=1
+#USE_BYTES=1
 #WARN_RX=0
 #WARN_TX=0
 #CRIT_RX=0
 #CRIT_TX=0
+#USE_SI=0
 ```
 
 E.g.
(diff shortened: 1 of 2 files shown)


bandwidth2/bandwidth2.c
```diff
@@ -142,21 +153,24 @@ int main(int argc, char *argv[])
...
+    warningtx = atoi(envvar);
   envvar = getenv("CRIT_RX");
   if (envvar)
-      criticalrx = atoi(envvar);
+    criticalrx = atoi(envvar);
   envvar = getenv("CRIT_TX");
   if (envvar)
-      criticaltx = atoi(envvar);
+    criticaltx = atoi(envvar);
+  envvar = getenv("SI");
```
kb100: Please use `USE_SI`

bandwidth2/bandwidth2.c
```diff
@@ -77,27 +78,36 @@ void get_values(char **const ifaces, int num_ifaces, time_t * const s, ulli * co
...
 
-void display(int const unit, double b, int const warning, int const critical)
+void display(int const unit, int const divisor,
+             double b, int const warning, int const critical)
 {
+  char fmtstr[7];
+
   if (critical != 0 && b > critical) {
-    printf("<span color='%s'>", RED);
+    printf("<span fallback='true' color='%s'>", RED);
```
kb100: I'm not familiar with the fallback option for pango, what does this do?

bandwidth2/README.md
```diff
@@ -25,11 +26,12 @@ markup=pango
 LABEL=
 #INTERFACE=(checks all interfaces)
 #USE_BITS=0
-#USE_BYES=1
+#USE_BYTES=1
 #WARN_RX=0
 #WARN_TX=0
 #CRIT_RX=0
 #CRIT_TX=0
+#SI=0
```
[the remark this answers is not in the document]
nolanl: done.

bandwidth2/bandwidth2.c
nolanl: done.

bandwidth2/bandwidth2.c
nolanl: Apparently allows fallback to other fonts if the specified font doesn't support a given unicode code point. I think it may have fixed it because I use a font-awesome symbol as a LABEL, and perhaps font-awesome doesn't have a code point for space (0x20)? Not entirely sure.

PR #210, opened 2019-02-08
volume-pulseaudio: Add option to output JSON

vladvasiliu:
Added an option to format the output of volume-pulseaudio as JSON. This allows the usage of both long and short formats with `SUBSCRIBE=1`. This is useful to keep the mute / unmute colors working.

### DIFF
--- a/volume-pulseaudio/volume-pulseaudio
+++ b/volume-pulseaudio/volume-pulseaudio
@@ -22,19 +22,17 @@ LONG_FORMAT=${LONG_FORMAT:-'${SYMB} ${VOL}% [${INDEX}:${NAME}]'}
 SHORT_FORMAT=${SHORT_FORMAT:-'${SYMB} ${VOL}% [${INDEX}]'}
 USE_ALSA_NAME=${USE_ALSA_NAME:-0}
 USE_DESCRIPTION=${USE_DESCRIPTION:-0}
-JSON_FORMAT=${JSON_FORMAT:-0}
 
 SUBSCRIBE=${SUBSCRIBE:-0}
 
 MIXER=${MIXER:-""}
 SCONTROL=${SCONTROL:-""}
 
-while getopts F:Sf:jadH:M:L:X:T:t:C:c:i:m:s:h opt; do
+while getopts F:Sf:adH:M:L:X:T:t:C:c:i:m:s:h opt; do
     case "$opt" in
         S) SUBSCRIBE=1 ;;
         F) LONG_FORMAT="$OPTARG" ;;
         f) SHORT_FORMAT="$OPTARG" ;;
-        j) JSON_FORMAT=1 ;;
         a) USE_ALSA_NAME=1 ;;
         d) USE_DESCRIPTION=1 ;;
         H) AUDIO_HIGH_SYMBOL="$OPTARG" ;;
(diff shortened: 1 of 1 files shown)


volume-pulseaudio/volume-pulseaudio
```diff
@@ -22,17 +22,19 @@ LONG_FORMAT=${LONG_FORMAT:-'${SYMB} ${VOL}% [${INDEX}:${NAME}]'}
 SHORT_FORMAT=${SHORT_FORMAT:-'${SYMB} ${VOL}% [${INDEX}]'}
 USE_ALSA_NAME=${USE_ALSA_NAME:-0}
 USE_DESCRIPTION=${USE_DESCRIPTION:-0}
+JSON_FORMAT=${JSON_FORMAT:-0}
```
kb100: i3blocks already has a `format` environmental variable builtin. Most blocks now use `format=pango` but in your case you  would put `format=json` in the i3blocks.conf file. Please use this `format` instead of creating a new one.

volume-pulseaudio/volume-pulseaudio
```diff
@@ -117,6 +120,11 @@ function print_format {
     echo "$1" | envsubst '${SYMB}${VOL}${INDEX}${NAME}'
 }
 
+function print_json {
+    echo "'{\"full_text\": \"$(print_format "$1")\", \"short_text\": \"$(print_format "$2")\", \"color\": \"$3\"}'"
```
kb100: Json has special characters that need to be escaped. For instance, if the user's format string has a `"` in it, then this will likely crash the block. The way to solve this is with by writing a `json_escape` function that takes text and escapes it so that special characters are escaped, e.g. all `"` are replaced with `\"` and so on.

volume-pulseaudio/volume-pulseaudio
vladvasiliu: If I understand the docs corectly, pango should go in `markup` not `format`. I've moved colour to pango.

volume-pulseaudio/volume-pulseaudio
vladvasiliu: I understand that random strings should be escaped, but having a proper parser in shell seems a bit too involved to me, especially since I don't believe the likelihood of having quotes coming from amixer is that high. Granted, users can configure the layout as they like, but again, getting quotes seems unlikely.

Anyway, I've removed all json formatting and I'll look into a different way of handling this, especially since in subscribe mode the blocklet doesn't get the events from the bar.

volume-pulseaudio/volume-pulseaudio
kb100: You are correct, pango is a markup thing not a format thing. However, from the man page of i3blocks:

     format This property specifies the format of the output text. The default format is plain text, as described in the  COMMAND  sec‐
                  tion. If "json" (or 1) is used, the block output is parsed as JSON.

The `format` key is a real key already part of i3blocks and `format=json` is the option that should be used in this case.

volume-pulseaudio/volume-pulseaudio
kb100: Actually I think this was recently added that persistent blocks now do receive click events. This PR no longer does what the title claims, so I'm closing it. Feel free to make a new PR with your new solution.

PR #211, opened 2019-02-09
mediaplayer: Output newline at end

Aster89:


kb100:
Looks good to me. I don't use this block so can you confirm that you tested it?

Aster89:
I confirm it works for `playerctl`. Since all the alternatives should be coherent (in terms of missing newlines), I'd say it just work anyway.

kb100:
Great, thanks for this one!

Aster89:
Just out of curiosity, what should the output be when Spotify (or whatever) is not running?

kb100:
I think the block should be transparent if the player is not running.

Issue #212, opened 2019-02-09
keyindicator's hardcoded output strings


Aster89:
This is not a bug report.

I understand that the only two possible values of the `KEY` variable (`CAPS` and `NUM`) are hardcoded,  but they are used straight away by `printf`, so that the output strings result hardcoded as well.

What about defining other two string variables for the texts relative to the two keys? (So that one can put in it whatever he wants, in terms of locale, symbols, ...).

kb100:
This is a fine idea if you would like to implement it, like an `ON_FORMAT` and `OFF_FORMAT` with defaults set to the `instance` variable.

Aster89:
I'll give a look at it tomorrow.

PR #213, opened 2019-02-10
keyindicator: customizable ON/OFF strings

Aster89:
To address my own proposal in #212.


kb100:
Please make the default behavior to do exactly what the current script does so we don't break anyone's bar.

kb100:
Meaning the defaults should not have fontawesome lock icons, and the default config also does not use locks icons.

Aster89:
Doesn't it? Each of the two variables I've added default to `BLOCK_INSTANCE` or `KEY`, if they're not provided in the i3blocks configuration file.

Aster89:
I've commented the two configuration lines, if that is what you implied.

Aster89:
@kb100, I think it should be ok, now, no?

da-edra:
Hello @Aster89 I know it's been a while since you opened this PR, I'd like to integrate it into a fork we're creating to keep the [i3blocks-contrib](https://github.com/da-edra/i3blocks-contrib) repository updated.

You PR looks pretty good! It is a nice and useful block, in case you're still using i3blocks and interested in it getting merged could you please fix the following errors [perl critic](https://metacpan.org/release/Perl-Critic) found?

```bash
Bareword file handle opened at line 64, column 1.  See pages 202,204 of PBP.  (Severity: 5)
Two-argument "open" used at line 64, column 1.  See page 207 of PBP.  (Severity: 5)
```

Thanks in advance!

Aster89:
Hi @da-edra. Unfortunately my `perl` is not good enough to take this on, at the moment. That line existed before I worked on this PR, however, and my edits were only some copy and paste.

Aster89:
@da-edra , license added.

Issue #208, opened 2019-02-05
[Request] A stopwatch and timer


parthsharma1996:
Hi,
I wanted to have a stopwatch and a countdown timer in my `i3bar` so that I can track time for various things. After a lot of googling I couldn't really find an `i3blocks` script that could do that. 

The functionality I am looking for is something similar to the stopwatch/timer applet that appears when you google "stopwatch" (which I currently use for my purposes)

1. Does any such script already exist of which I am not aware? 
2. If not, I would like to request for such a script. I can also try and write my own in python if someone could give me a broad strokes idea on how to do it. 

Thanks!

Aster89:
I've been thinking about the very same blocklet, with same very hint (Google). The hard part, I think, is that the script should access the blocklet state when you click on it. E.g. when the timer is paused and you click on it to resume/start it, the script should first retrieve the current time, and then start running. And... Oh, I'm just teasing about it. Take what I've written so far for what it is 🤣

Aster89:
Other thought: you want the blocklet to be updated with a high frequency (at least once per second, like the clock, it quicker if you want the decimals). But you you don't want it to refresh (at all) when it's paused

kb100:
This is definitely possible if one of you would like to try your hand at it. You will need to use `format=json` and `interval=persist` and loop over a `read -t ` statement. Persistent blocks get click info through stdin so that's why you will need to be using `read`.

Aster89:
_You will need_? Is json a necessary condition to give shape to this idea? I ask this because the only json I can write something sensible about has one more letter in it and comes out of Crystal lake...

kb100:
Well, for the block to be fully functional I think you would need json, since normal and pango formatted blocks only receive the block button to stdin and no other information (click location, click modifiers, etc.). I guess a stopwatch could be fully functional with just the button, but I would imagine a timer block or combined timer/stopwatch to need more. If you come up with something nice that doesn't use json that's fine by me!

Aster89:
@kb100 , I think I'm finding a way :smiley:

PR #215, opened 2019-02-12
timer_and_stopwatch: minimal stopwatch and colored timer

Aster89:
@kb100, in response to issue [#208](https://github.com/vivien/i3blocks-contrib/issues/208), I coded this minimal stopwatch and timer blocklet. 

The timer has also a (hard-coded) coloring that goes linearly from green to red.

I've also added a variable to make the timer loop around. I think it's useful when you have, say, X questions you should answer in Y seconds; you can put the timer at Y/X and it will track the time per question.

Let me know if I can improve it.


da-edra:
Hello @Aster89 I know it's been a while since you opened this PR, I'd like to integrate it into a fork we're creating to keep the [i3blocks-contrib](https://github.com/da-edra/i3blocks-contrib) repository updated.

You PR looks pretty good! It is a nice and useful block, in case you're still using i3blocks and interested in it getting merged could you please fix the following errors [shellcheck](https://github.com/koalaman/shellcheck) found?

```bash
In timer.sh line 24:
    $running && pause || play
             ^-- SC2015: Note that A && B || C is not if-then-else. C may run when A is true.


In timer.sh line 64:
    hue360=$(( 120*$t/$initial_time ))
                   ^-- SC2004: $/${} is unnecessary on arithmetic variables.
                      ^-----------^ SC2004: $/${} is unnecessary on arithmetic variables.


In timer.sh line 98:
    for i in ${!modes[@]}; do
             ^----------^ SC2068: Double quote array expansions to avoid re-splitting elements.


In timer.sh line 156:
echo '{"full_text":"'$full_text'"' \
                     ^--------^ SC2086: Double quote to prevent globbing and word splitting.

Did you mean:
echo '{"full_text":"'"$full_text"'"' \


In timer.sh line 157:
     ',"status_symbol":"'$status_symbol'"' \
                         ^------------^ SC2086: Double quote to prevent globbing and word splitting.

Did you mean:
     ',"status_symbol":"'"$status_symbol"'"' \


In timer.sh line 162:
     ',"mode":"'$mode'"'    \
                ^---^ SC2086: Double quote to prevent globbing and word splitting.

Did you mean:
     ',"mode":"'"$mode"'"'    \

For more information:
  https://www.shellcheck.net/wiki/SC2068 -- Double quote array expansions to ...
  https://www.shellcheck.net/wiki/SC2015 -- Note that A && B || C is not if-t...
  https://www.shellcheck.net/wiki/SC2086 -- Double quote to prevent globbing ...
```

Thanks in advance!

Aster89:
Yes, I'm still using this `i3blocks` and I'll give a look at this blocklet asap.

Aster89:
Ok, @da-edra , so I've used `cat` with a "here document" in place of `echo`/`printf` for printing the JSON line so I could save several characters and quotes.

I've skipped only one of those `spellcheck` "suggestions", for this reason: both `play()` and `pause()` functions just set the variable `running` to `true`/`false` (and there's no way this can fail, is there?), and the variable `status_symbol` to a value to be set from `i3blocks`'s config file. This is to say that I can't see how, in this case, `A && B || C` could behave differently from _if-then-else_. Spelling out the latter would mean either 5 lines or a very ugly single line.

Aster89:
@da-edra , I think you might have erroneously commented the wrong PR. Maybe you ment to comment #213 ?

PR #218, opened 2019-02-22
metars: add metars, a script for METAR weather data

beoldhin:


kb100:
Hi, thanks for providing this. The bulk of the script looks really good. For configuration though, all scripts in this repository now use injected properties for configuration so that users only need to edit their i3blocks configuration, not the script itself, to customize the behavior. Would you be up to the task of modifying the metar script to do this?

beoldhin:
Injected properties are now implemented.

PR #230, opened 2019-04-19
aur-update: add support for LABEL

billux:
aur-update blocklet now displays the content of `LABEL` environment variable before its output:

```
$ LABEL=AUR aur-update/aur-update 
AUR<span color='yellow'>2 AUR updates</span>
[…]
```


jolange:
I don't think having the outptut "AUR2 AUR updates" is desirable. It should rather be "&lt;label>2 updates", such that `LABEL="AUR " aur-update/aur-update` woud give you "AUR 2 updates".

billux:
Actually I used it to have a Font Awesome glyph before the text. With LABEL= I have: output

jolange:
Ok, but for more general usage, I'd still suggest my version. Then you could use "LABEL= AUR" and would get  " AUR 2 updates".

jolange:
You have pushed a couple of unrelated commits to you master branch now, which is ok for your private fork, of course.
But you still have this PR open here (from your master branch to the upstream master branch). This means you update this PR with all your new commits (and we get notifications for that).

You should close this and if you want to have your changes merged, please open a separate PR with a new branch containing only the relevant changes.

PR #236, opened 2019-06-02
Update bandwidth2.c

brijgogogo:
fixed error: bandwidth2.c:97:3: note: ‘snprintf’ output 8 bytes into a destination of size 7

### DIFF
--- a/bandwidth2/bandwidth2.c
+++ b/bandwidth2/bandwidth2.c
@@ -94,7 +94,7 @@ void display(int const unit, int const divisor,
   if (unit == 'b')
     b = b * 8;
 
-  snprintf(fmtstr, sizeof (fmtstr), "%%%d.1lf ", divisor > 1000 ? 6 : 5);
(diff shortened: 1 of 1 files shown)


kuroneko9:
Also, on line 58

```
        if (strcmp(ifaces[i], ifname) != 0) {
```

should be

```
        if (strcmp(ifaces[i], ifname) == 0) {
```

kuroneko9:
Line 54:

```
    if (sscanf(line, "%[^:]: %llu %*u %*u %*u %*u %*u %*u %*u %llu", ifname, &temp_r, &temp_s) == 3) {	
```

could be

```
    if (sscanf(line, " %[^:]: %llu %*u %*u %*u %*u %*u %*u %*u %llu", ifname, &temp_r, &temp_s) == 3) {	
```

Because in `/proc/net/dev`, there can be spaces before interface names, adding a space to the format string before `%[^:]` will prevent those spaces to be included into `ifname`.

PR #240, opened 2019-06-23
battery icon changes based on battery percentage

ramenbroth:


lucassperez:
I did the exact same thing here lol\
Then I decided to open a PR for this change, but found yours. I wish they would merge this :sweat_smile:

PR #243, opened 2019-07-22
Update kbdd_layout

bergentroll:
Update a bit kbdd_layout script:
- Replace sed pipes with single grep matching;
- Add some flexibility in output format; 
- Remove tabs.


bergentroll:
Friends?

da-edra (approved):
Thank you so much! <3 
And sorry for the late approval and merge.

PR #262, opened 2019-09-25
[arch-update] Add ability to set any aur helper and yay as default

hashworks:
This additionally replaces yaourt with yay as the default. I think that's reasonable since yay is quite popular and yaourt is a persona non grata in arch (gone from the wiki and the AUR).

Please note that yaourt can still be used:
```
AUR_HELPER=yaourt
AUR_PREFIX=aur/
```


hashworks:
@kb100 is that acceptable?

PR #270, opened 2019-10-07
Update arch-update for change of checkupdates

LEXUGE:
It seems like that `checkupdates` will exit 1 if there is no update available, causing unexpected behavior of this script.
I revised it a little bit and it works fine now.

### DIFF
--- a/arch-update/arch-update
+++ b/arch-update/arch-update
@@ -62,7 +62,14 @@ def create_argparse():
 
 
 def get_updates():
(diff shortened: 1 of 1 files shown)


jolange:
Thanks!

Merged as 19fe4bdf04fd7c0cb40c31e870befaeeb8a882f0.
Additionally, I changed it to exit code 2, [because that's what it returns by now](https://git.archlinux.org/pacman-contrib.git/commit/?id=3da550ef93ae0b999836181044d33a8147a00088).

PR #277, opened 2019-11-01
wifi: simplify and reduce dependencies

The-King-of-Toasters:
The script is rewritten to use standard shell and awk with bashisms,
shell variables and extraneous lines removed. Wi-Fi quality extraction
and color handling has been moved into awk as well. The README has been
tweaked slightly to add syntax highlighting and other markup
improvements.

### DIFF
--- a/wifi/README.md
+++ b/wifi/README.md
@@ -1,13 +1,13 @@
-# wifi
+# `wifi`
 
-Show the strength of your wifi connection.
-If no instance is specified, wlan0 is used.
+Show the strength of your Wi-Fi connection.
+If no interface is specified, `wlan0` is used.
 
 ![](wifi.png)
 
-# Config
+## Config
 
-```
+```ini
 [wifi]
 command=$SCRIPT_DIR/wifi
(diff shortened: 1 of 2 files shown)


Aster89:
@The-King-of-Toasters , maybe you could make more settings customizable through variables, like the color and/or text when the interface is down; I might want to have "assente", which is italian, rather than "down", or I might prefer grey over red. What do you think about it?

The-King-of-Toasters:
I can make that work, it'll take a bit.

The-King-of-Toasters:
Actually, I take that back. If you'd like to modify the script I recommend downloading/forking it yourself. As of now I'll stick to what was already written, perhaps adding the code for #225.

@kb100 @vivien Any thoughts?

Aster89:
@The-King-of-Toasters , I can certainly do it, but can I ask what keeps you from doing it? I mean, I could agree with you and renounce myself.

Concerning the PRs, it looks to me that all has been steady for months. Do you know anything about it?

Issue #284, opened 2019-11-20
arch-update: handle error exit of checkupdates


uykusuz:
`checkupdates` v1.2.0 exits with exit code `1` when there are no updates. 

### Expected behavior

`arch-update` should return that the system is up-to-date when `checkupdates` exits with an error. 

### Actual behavior

`arch-update` fails and the whole blocklet won't get displayed.

### i3blocks config relevant to blocklet(s)

arch-update

### Output of blocklet(s) when run from command line

```sh
Traceback (most recent call last):
  File "./arch-update", line 109, in <module>
    updates = get_updates()
  File "./arch-update", line 65, in get_updates
    output = check_output(['checkupdates']).decode('utf-8')
  File "/usr/lib/python3.7/subprocess.py", line 395, in check_output
    **kwargs).stdout
  File "/usr/lib/python3.7/subprocess.py", line 487, in run
    output=stdout, stderr=stderr)
subprocess.CalledProcessError: Command '['checkupdates']' returned non-zero exit status 1.
```

### Output of any relevant other commands that might help diagnostics

`checkupdates` exits with `1` when there are no updates.

jolange:
Thanks, it's fixed in f2715bc549453efaaf3a4998a9b6b418a3f044d4 now.

PR #285, opened 2019-11-20
arch-update: handle non-zero exit code of checkupdates

uykusuz:
checkupdates v1.2.0 exits with exit code 1 when there are no updates.
In that case the script should not fail but report "system up to date".

### DIFF
--- a/arch-update/arch-update
+++ b/arch-update/arch-update
@@ -62,7 +62,14 @@ def create_argparse():
 
 
 def get_updates():
-    output = check_output(['checkupdates']).decode('utf-8')
+    output = ''
+    try:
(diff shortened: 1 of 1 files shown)


uykusuz:
This PR is for issue #284

MarcoMiano:
in checkupdates 1.3.0-1 (from pacman-contrib 1.3.0-1) checkupdates exti with 2 when there are no updates.
see my comment on #294

uykusuz:
@MarcoMiano added handling of checkupdates v 1.3

jolange:
Thanks, it's fixed in f2715bc549453efaaf3a4998a9b6b418a3f044d4 now.

PR #287, opened 2019-12-09
updates volume section in i3blocks.conf

sumeet221b:
capital label was not working


jolange:
You are probably using an outdated version of i3blocks.
See https://github.com/vivien/i3blocks/issues/279 and #132.

PR #290, opened 2019-12-11
Arch update: add yay

uykusuz:
This will simply try to get updates from yaourt and if it fails try yay.

### DIFF
--- a/arch-update/arch-update
+++ b/arch-update/arch-update
@@ -62,7 +62,14 @@ def create_argparse():
 
 
 def get_updates():
-    output = check_output(['checkupdates']).decode('utf-8')
+    output = ''
+    try:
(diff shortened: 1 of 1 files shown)


arch-update/arch-update
```diff
@@ -62,7 +62,14 @@ def create_argparse():
 
 
 def get_updates():
-    output = check_output(['checkupdates']).decode('utf-8')
+    output = ''
+    try:
+        output = check_output(['checkupdates']).decode('utf-8')
+    except subprocess.CalledProcessError as exc:
+        # usually this means no updates are available
+        if not (exc.returncode == 1 and not exc.output):
```
treeshateorcs: the return code for no updates is 2

PR #292, opened 2019-12-25
Fix extra newline

mgbowe1:
Fixes bug I found when attempting to use the mediaplayer blocklet.
Remove extra newline returned by playerctl artist metatdata that was preventing track title from showing by default


mgbowe1:
The Artist field returned when spotify was the player had a newline character on the end, so any track playing would only show the artist until a call to strip the newline was added.

PR #295, opened 2020-01-07
New block: ssid:  SSID of current wifi connection

hseg:
Based on `iface` script. Uses `iw` instead of `iwconfig`, which is deprecated
according to <https://wireless.wiki.kernel.org/en/users/documentation/iw>.
As well, avoids the perl dependency, though that's not much of a benefit.


da-edra:
Hello @hseg  !
I now have write access to this repository so I'll be merging your branch, thanks for sharing! :1st_place_medal:

PR #298, opened 2020-01-16
mediaplayer: Added play/pause icons for playerctl, updated screenshot.

DylanAuty:
Using the playerctl status command to output unicode play or pause icons according to the current status, prepended to the current song information.


mediaplayer/mediaplayer
```diff
@@ -119,9 +119,24 @@ sub mpd {
...
 
+    my $playerStatus = qx(playerctl $player_arg status);
+    chomp $playerStatus;
+    # exit status will be nonzero when playerctl cannot find your player
+    exit(0) if $? || (($playerStatus ne 'Playing') && ($playerStatus ne 'Paused'));
+
+    my $playPauseIcon = '  ';
+    if ($playerStatus) {
+        if ($playerStatus eq 'Playing') {
+            $playPauseIcon = '  ';
```
adhikasp, line 130:
```suggestion
            $playPauseIcon = '▶  ';
```
https://www.fileformat.info/info/unicode/char/25b6/index.htm

mediaplayer/mediaplayer
```diff
@@ -119,9 +119,24 @@ sub mpd {
...
+        } elsif ($playerStatus eq 'Paused') {
+            $playPauseIcon = '  ';
```
adhikasp, line 132:
```suggestion
            $playPauseIcon = '⏸  ';
```
https://www.fileformat.info/info/unicode/char/23f8/index.htm

DylanAuty:
I was using fontawesome, I think, and the unicode I went with was based on the ones they gave on their site. I didn't think this would be an issue, sorry about that! I've rolled the changed you made into a commit, thank you for reviewing this.

Issue #255, opened 2019-09-14
config.example fix command path


LukasLeppich:
### Expected behavior
The config.example should work if i change $SCRIPT_DIR to the location of this repository.

### Actual behavior
```
Error message: /bin/sh: /home/{username}/.config/i3blocks/volume: Is a directory
...
```

### i3blocks config relevant to blocklet(s)
config.example from this repository.
Relevant line:
[command=$SCRIPT_DIR/$BLOCK_NAME](https://github.com/vivien/i3blocks-contrib/blob/master/config.example#L28)

### Output of blocklet(s) when run from command line
```
/bin/sh: /home/{username}/.config/i3blocks/volume: Is a directory
<3>[volume] Command '/home/{username}/.config/i3blocks/$BLOCK_NAME' not executable
...

```

### Works for me:
If i change the line to `command=/home/{username}/.config/i3blocks/$BLOCK_NAME/$BLOCK_NAME` it works. 

The original line executes /.config/i3blocks/volume which is an directory. If i change it to $BLOCK_NAME/$BLOCK_NAME it executes the script /.config/i3blocks/volume/volume.

skidnik:
From CONTRIBUTING.md:

> A blocklet MUST be confined to a single directory and the name of the directory MUST be relevant to the blocklet's purpose or core feature.

> The command file's name, SHOULD match the name of the containing directory.

This does not guarantee that the blocklet directory name matches the blocklet executable name. Thus locking both subdirectory and executable name to the same variable is wrong.

In addition, the the right place to store blocklets in use is under 'libexec' directory, e.g. `/usr/libexec/i3blocks` or `${HOME}/.local/libexec/i3blocks`. They have unique names, so nothing prevents you from storing them under the same directory.

End users on the other hand can do whatever is to their desire, but that, as in you case, requires manual intervention.

therealmaxmoon:
I had the same issue, that only the date was seen with the config.example. Changing the command line to
`command=$SCRIPT_DIR/$BLOCK_NAME/$BLOCK_NAME`
helped. But it isn't guaranteed that the directory name is the same like the name of the plugin.

@skidnik why exactly /usr/libexed/? Why not ~/i3blocks/i3blocks-contrib-executables or any other directory in $HOME? 
And is there a way to copy all the executable faster than looking for them by hand?

skidnik:
@therealmaxmoon 

> why exactly /usr/libexe?

[...]

PR #307, opened 2020-02-24
Coerce to current folder structure

blaggacao:
Each executable currently is placed in a folder with the same name.

### DIFF
--- a/config.example
+++ b/config.example
@@ -25,7 +25,7 @@
 # The top properties below are applied to every block, but can be overridden.
 # Each block command defaults to the script name to avoid boilerplate.
 # Change $SCRIPT_DIR to the location of your scripts!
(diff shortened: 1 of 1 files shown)


skidnik:
CONTRIBUTING.md:

> The command file's name, SHOULD match the name of the containing directory. E.g. myscript/myscript is a good name, but myscript/yourscript is not.

[RFC2119](https://www.ietf.org/rfc/rfc2119.txt)

Unless contributing rules are also changed, this is wrong.

blaggacao:
@skidnik Is yours an implicit argument _for_ this PR?

skidnik:
Yes, current contributing rules allow block directroy to not match block script name.

#255 issue on the same question is still open.

PR #311, opened 2020-02-26
Add option to memory to allow not showing the percentage

someonewithpc:


zoj613:
Would love for this feature to be added. I find the percentage output not necessary in most cases.

PR #315, opened 2020-03-18
iface/bandwidth: fix display when using VPN

kuepe-sl:
The scripts rely on "ip route" output of the format
  default via IP dev INTERFACE ...
However, when using VPNs the output may be
  default dev INTERFACE ...

The additional sed command removes everything between "default" and
"dev" in order to fix this issue.


yutakatay:
any update on this PR?

@jolange @da-edra 
I'd like to check this PR.

PR #319, opened 2020-03-28
Catch exception when checkupdates exits with statuscode 2

vodchella:
Catch exception when checkupdates exits with statuscode 2 (system up to date)


jolange:
Thanks, it's fixed in f2715bc549453efaaf3a4998a9b6b418a3f044d4 now.

PR #338, opened 2020-05-23
Fix strcmp comparison

pcaro90:
`strcmp` returns `0` when two strings are equal, which is what the `if` should be cheking. Instead, the fixed comparison was looking for strings which were different to the given interface.


pcaro90:
Also, short interface names appear with leading whitespaces in `/proc/net/dev`, causing the `sscanf` function to not match. By adding a leading space in the format string, `sscanf` matches zero or more whitespace characters.

PR #339, opened 2020-06-03
gpu-load: support AMD based card

adhikasp:
This add support for showing AMD based card with `radeontop` dependency.

Note that it only support 2 of 4 existing metrics because `radeontop` doesn't expose the same variable as `nvidia-settings`.

image


da-edra (approved):
Hello @adhikasp ! 👋 

This is an **awesome** enhancement! 😍 

Thank you so much for your pull request! 🤗

PR #345, opened 2020-06-27
[battery2] Make colors configurable

m8mble:
Honor user overwrites, fallback to previously hard-coded colors.


da-edra (approved):
Awesome!! Thank you so much!! 🥇

PR #356, opened 2020-09-03
miccontrol: add default miccontrol blocklet

fpaskali:


da-edra (approved):
Thank you very much for sharing your blocklet with us! 🥇

PR #358, opened 2020-09-13
Add battery-upower

lunaryorn:
Show battery status from [UPower] daemon.

battery-upower

The blocklet shows the power plug character 🔌 if the battery is charging or fully charged, or the battery character 🔋otherwise, along with the time to full or empty respectively.  If UPower provides no information about the remaining time show the battery level in percentage instead.

As short text the blocklet shows the percentage along with a `+` sign (charging), a `-` sign (discharging) or no sign (fully charged).

This blocklet does not poll the battery; instead it is a persisted blocket which connects to the UPower DBus interface at startup and listens for change signals emitted by UPower.

[upower]: https://upower.freedesktop.org/


troyready:
Looks really great. Any chance this could be adjusted to account for systems with multiple batteries? A number of Lenovo systems have a configuration like this:

```
upower.EnumerateDevices()
['/org/freedesktop/UPower/devices/line_power_AC', '/org/freedesktop/UPower/devices/battery_BAT0', '/org/freedesktop/UPower/devices/battery_BAT1']
```
In this case, the battery percentage should be the average of the percentage of the 2 batteries.

lunaryorn:
Sure you just need to do it.  The code's there for you to edit.

troyready:
I'd be open to PRing your fork, but GitHub is showing the source of this PR as an `unknown repository` -- can you restore your fork & branch?

lunaryorn:
No, I don't really care for this pull request.  You can perhaps clone it directly: the for also exists in my dotfiles repo.

battery-upower/battery-upower
```diff
@@ -0,0 +1,133 @@
...
+
+def find_battery(upower):
+    for device_path in upower.EnumerateDevices():
+        device = Gio.DBusProxy.new_sync(
+            upower.get_connection(), Gio.DBusProxyFlags.NONE,
+            None, upower.get_name(), device_path,
+            'org.freedesktop.UPower.Device')
+        if device.get_cached_property('Type').unpack() == DeviceType.BATTERY:
+            return device
+    else:
```
da-edra, line 50: I think you don't need a `for/else` loop here because on line 49 you have a `return` statement that would continue the execution of the code in the `main()` function. The `for/else` loop is useful for when you have a `break` statement in your `for` loop because the execution of the code would continue in the same function where the `for` loop is declared. This is not the case.

I'd suggest changing the `for/else` for a `for` loop and returning at the end, like this:

```python3
def find_battery(upower):
    for device_path in upower.EnumerateDevices():
        device = Gio.DBusProxy.new_sync(
            upower.get_connection(), Gio.DBusProxyFlags.NONE,
            None, upower.get_name(), device_path,
            'org.freedesktop.UPower.Device')
        if device.get_cached_property('Type').unpack() == DeviceType.BATTERY:
            return device
    return None
```

What do you think?

da-edra (changes requested):
Hello @lunaryorn ! 👋 

Thank you so much for sharing your `blocklet` with us 🤗 , it looks great I just left a comment for a small code improvement.

Please let me know if you have any doubts or suggestions regarding it. 😉

lunaryorn:
@da-edra Thanks but as I said

> I don't really care for this pull request.

da-edra:
Okay, then I'll close it.

PR #359, opened 2020-09-14
iface: output wifi name if possible

uykusuz:
If iw is present on the system, it will try to find the wifi name and display that as short text.
As long text the IP and the wifi name will be shown.
I removed the spaces after the label. This aligns with all other blocks.
I.e. that you don't have to set LABEL=mywifi<space>

### DIFF
--- a/iface/iface
+++ b/iface/iface
@@ -43,16 +43,16 @@ for flag in "$1" "$2"; do
       AF=inet6 ;;
     -L)
       if [[ "$IF" = "" ]]; then
-        LABEL="iface "
+        LABEL="iface"
       else
-        LABEL="$IF: "
+        LABEL="$IF:"
       fi ;;
   esac
 done
 
 if [[ "$IF" = "" ]] || [[ "$(cat /sys/class/net/$IF/operstate)" = 'down' ]]; then
-  echo "${LABEL}down" # full text
-  echo "${LABEL}down" # short text
+  echo "${LABEL} down" # full text
+  echo "${LABEL} down" # short text
   echo \#FF0000 # color
   exit
 fi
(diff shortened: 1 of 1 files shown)


da-edra (approved):
Thanks a lot! <3

atreyasha:
With this commit I no longer see my IP address but only see my wifi name. Would be good if the user could optionally choose this behaviour through the `conf` file.

uykusuz:
@atreyasha oh right, just to make sure: it's not enough to see your IP when you click on the wifi name?

atreyasha:
@uykusuz ah yes that's right, when I click I can see the wifi name and the IP address.

Just my two cents: I would have preferred an environmental variable (eg. `DISPLAY_WIFI_NAME`) that activates this behaviour and can be set directly in the `i3blocks.conf` file, something along the lines of:

```shell
if [ "$DISPLAY_WIFI_NAME" = "1" ]; then
  if command -v iw > /dev/null && iw $IF info > /dev/null 2>&1;
  then
      WIFI_NAME=$(iw $IF info | grep -Po '(?<=ssid ).*' | tr -d " \t\n\r")

      if [[ $BLOCK_BUTTON -eq 1 ]];
      then
          message="$LABEL $WIFI_NAME ($IPADDR)"
      else
          message="$LABEL $WIFI_NAME"
      fi
  else
      message="$LABEL $IPADDR"
  fi
  echo "$message"
else
  echo "$LABEL$IPADDR" # full text
  echo "$LABEL$IPADDR" # short text
fi
```

uykusuz:
@atreyasha please review https://github.com/vivien/i3blocks-contrib/pull/402

PR #364, opened 2020-10-22
wifi: exposing colors as variables

Aster89:
One can change the colors of the i3 bar, therefore some colors of the applets can be inappropriate (e.g. i3 color is green? Green is then not a good color to say the WiFi has 100% signal), therefore exporting the colors makes thing easier for the user. This is true in general for all applets, actually, but I'm addressing only one.

I've kept the originally hardcoded colors as the defaults for the variables I'm exposing.


da-edra (approved):
Thank you!

PR #370, opened 2020-11-22
screen-recorder: Record your desktop session to an animated GIF

karta0807913:


karta0807913:
# screen-recorder

Record your desktop session to an animated GIF.

The recorded file will be opened in the browser.





# Requirements

- `byzanz-record`,
- `x-www-browser`,
- `nohup`
- `slop`

# Usage

```ini
[screen-recorder]
interval=once
```

Aster89:
I suspect you have not recorded the gif showing your recording blocklet by using your recording blocklet, right? This seems a very useful blocklet, however!

karta0807913:
@Aster89 thank you for reply.

Let me explain this.

In command `byzanz-record`, there is a `-e` option which fork another process to execute and time. 
Then use `tail --follow name` to follow the file which will exit when file been delete.

Therefore, we can do some tricks, such as copying the script to another location to record at the same time.



```
out1.gif
```



And this one is recorded from bar.

PR #371, opened 2020-11-23
timer_and_stopwatch: add UNIX_TIME parameter to fix fullscreen use

stephengmatthews:
When LibreOffice Impress runs in presentation mode (press `F5`), the timer/stopwatch is _not_ updated. The timer/stopwatch is updated when using full screen mode in i3, such as Firefox and terminal.

This PR relates to [#255](https://github.com/vivien/i3blocks/issues/255) in i3blocks (not i3blocks-contrib). I tried to find the cause of the issue but I admit I didn't know where to look (i3blocks? LibreOffice?). For this reason, I've added this as a config parameter to allow enabling/disabling rather than pushing for a bug fix.

This PR adds a UNIX_TIME parameter to the config. When enabled, unix time is used to calculate the time difference for incrementing the timer/stopwatch. The idea is that if the `timer_and_stopwatch` script is not run every second, say for 10 seconds because LibreOffice Impress is in presentation mode, then the next time the script runs the difference in unix time is 10 seconds, which is used for incrementing the timer/stopwatch. I've tested it and it works for me.

`UNIX_TIME` defaults to false, but can be enabled with `UNIX_TIME=true` in the config.

What are your thoughts?

### DIFF
--- a/timer_and_stopwatch/README.md
+++ b/timer_and_stopwatch/README.md
@@ -25,4 +25,5 @@ interval=1
 #PLAY_LABEL=(playing)
 #PAUSE_LABEL=(paused)
 #TIMER_LOOP=true
+#UNIX_TIME=true
 ```
--- a/timer_and_stopwatch/timer_and_stopwatch
+++ b/timer_and_stopwatch/timer_and_stopwatch
@@ -8,6 +8,7 @@ DEFAULT_TIMER=${DEFAULT_TIMER:-60}
 PLAY_LABEL=${PLAY_LABEL:-(playing)}
 PAUSE_LABEL=${PAUSE_LABEL:-(paused)}
 TIMER_LOOP=${TIMER_LOOP:-false}
+UNIX_TIME=${UNIX_TIME:-false}
 ###### Default environment variables ######
 
 ###### Functions ######
@@ -40,6 +41,7 @@ reset_times() {
 }
 
 set_mode() {
+  previous_unix_time=-1
   case ${modes[$mode]} in
     'timer' )
       running=false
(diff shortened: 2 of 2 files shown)


Aster89:
Wow! I can't believe someone is really using my blocklet 🤭. I'll give a look at this pr myself later, even if I don't have any permission to accept/deny/whatever.

stephengmatthews:
> Wow! I can't believe someone is really using my blocklet. I'll give a look at this pr myself later, even if I don't have any permission to accept/deny/whatever.

:) It's a useful blocklet -- thank you! I use it to track my time when working.

Aster89:
Feeback on the _reason_ for this change. Without it, I can start the timer, go full screen in Impress with <kbd>F5</kbd>, wait a few second, exit full screen, and see that the time _has_ elapsed.

Probably the behavior you observe depends on something other than i3/i3blocks, because I use them both.

stephengmatthews:
Running i3blocks with `-vv`, I see `sigwaitinfo(): Interrupted system call` on standard error at some point between when Impress' full screen starts and stops. I don't know how to investigate further. Would the i3blocks repo be the correct place to raise this?

Aster89:
Why not? Worst case scenario: someone will tell you what is he correct place ;)

stephengmatthews:
I queried this behaviour on the i3blocks repo and they responded [here](https://github.com/vivien/i3blocks/issues/440#issuecomment-733980638). It seems that the timer/stopwatch not updating is expected.

@Aster89, you tried it for "a few second", would you mind trying for 5-10 seconds, which is how long I waited? I started using 1/2/3 seconds but sometimes found it difficult to spot non-updates after a few seconds. Also, I noticed the non-updates occur only when all i3bars are obscured when multiple monitors are used.

Aster89:
@stephengmatthews, I've rectified my comment, as I did wait enough. Furthermore I've just tried now also on my laptop (same configuration: archlinux, i3, i3blocks, ...), and the timer seems to go just fine: I counted 20 seconds right after pressing <kbd>F5</kbd>, then pressed <kbd>Escape</kbd>, and the timer correctly showed 20 seconds less.

stephengmatthews:
Thanks for trying again @Aster89.

I've tried on my laptop without external monitor, and I see the timer/stopwatch not updating -- so, same as before for me.

I used the same empty presentation as yourself.

To test and poke around a bit more, I set the [display mode](https://i3wm.org/docs/userguide.html#_display_mode) to `hide`, restarted i3, and verified that: 1) the timer/stopwatch did not update when i3bar was hidden, and 2) the timer/stopwatch did update whilst holding the mod key. Could you test this setting, please? I'm hoping this could reveal something to point us in the right direction!

stephengmatthews:
Forgot to mention: I'm using Ubuntu 18.04.

Aster89:
With the bar mode set to hide, I observe the same behavior you describe, but this has nothing to do with LibreOffice Impress. Maybe you can try address this, and you'll find the issue with impress solved automatically, I don't know.

In comparison, the following blocklet for time and date works just fine
```
[time]
command=date '+ %H:%M:%S'
interval=1
separator=false
```

PR #372, opened 2020-11-23
timer_and_stopwatch: export neutral background color

Aster89:
I am not using this blocklet much, but I've used it again and realized I hardcoded `#000000` (black) as the color for the neutral background.

For a while however, I've been using a non-black i3 bar, and having the background of a blocklet black is really ugly.

This commit exposes the variable `NEUTRAL_COLOR` to specify that color; it defaults to `#000000` otherwise.

As regards the other colors (for instance the colors from green to red for the timer), which can also be not ok for users who decide to give their i3 bars an "unconvetional" coloring, I think I could in the future export something like initial color, final color, and similar.

### DIFF
--- a/timer_and_stopwatch/README.md
+++ b/timer_and_stopwatch/README.md
@@ -1,8 +1,8 @@
 # timer_and_stopwatch
 
-Have a minimal timer and stopwatch blocklet. Right click to switches between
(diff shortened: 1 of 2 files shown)


da-edra (approved):
Thanks a lot for your awesome contributions @Aster89 !! 🥇 🙇‍♀️

PR #373, opened 2020-12-12
disk-io: Allow user-specified alignment

BourgondAries:
Allows the read part to be aligned by the user. Set `ALIGN=` to do so.


da-edra (approved):
Thanks!

PR #377, opened 2021-01-10
memory2: created

trobol:
Simple bar displaying memory usage.


da-edra (approved):
Thank you so much! ❤️

PR #379, opened 2021-01-13
ssid: fix for ssids with spaces

mads256h:
This fixes a problem with ssids with spaces.
If the ssid is named "My network" the script would only print "My".
I have fixed this so it now prints "My network".


da-edra (approved):
Thank you so much!

PR #382, opened 2021-01-25
Fix shebang

avitase:
Add missing / in shebang


da-edra:
Thank you! This PR closes #386

PR #384, opened 2021-01-29
Add optimus-manager block

markx3:
Hello!

In this PR we add a simple block that displays/switches the current active GPU. 

Its dependencies are:
- `optimus-manager` (so, NVIDIA laptops only).


da-edra (approved):
Thank you!

PR #385, opened 2021-02-20
nm-vpn: fix vpn naming

achuhlomin:
In case there is no VPN connection type, but TUN connection type established, naming is broken (e.g. ProtonVPN works so). So we can just use TUN name in that case VPN name absent.


da-edra (approved):
Thank you!

PR #387, opened 2021-03-17
usb: avoid parsing lsblk's pretty-printing

hseg:
Also pep8 code, cleanup


da-edra (approved):
Thank you so much! <3
 
I really like the approach you took to solve giant chunk of code we had in the `def parseArguments` function. Your approach is pretty nice, clean and follows the D. R. Y. Principles. 🥇

PR #390, opened 2021-03-29
rofi-calendar: add BAR_POSITION and WEEK_START variables

xinye1:
All changes are non-breaking, as default values for new variable are set. Details:

1. `BAR_POSITION` is added to make calendar visible for a bar on the top of the screen. This is done by specifying `anchor` variable for `rofi` command passed by the added variable. Possible further improvement is possibly passing a `$BAR_POSITION` variable from i3's `config` to `i3block.conf`.
2. `WEEK_START` simply takes advantage of the `--monday` and `--sunday` arguments of `cal` command.

`README.md` is updated to reflect the above.


da-edra (approved):
谢谢！😊👌🏽

PR #393, opened 2021-04-27
brightness: non blocking brightness monitoring

Lezh1k:
1. Written in C
2. Using files which are used by brightnessctl.


da-edra (approved):
Thank you so much! 😊

PR #394, opened 2021-04-28
backlight: remove decimal places from displayed screen brightness

pabloariasal:
`xbacklight -get` (v1.2.3) shows several decimal places:

image

Most users aren't interested in subdecimal screen brightness, but want a rounded value:

image


da-edra (approved):
Thanks! 😊

PR #396, opened 2021-05-03
Switch to dunstctl because of deprecation warnings

deimosfr:
In order to follow dunst changes, dunsctl should be used now


da-edra (approved):
Thanks.

PR #397, opened 2021-05-09
purpleair: add purpleair, reports air quality values from PurpleAir

majorde1:
This is my first pull request into this project. I believe that it satisfies all of the contributing guidelines, but apologies if I missed something.


da-edra (approved):
Awesome! Thank you so much for sharing your block with us! <3

PR #398, opened 2021-05-10
timer_and_stopwatch: conditional expression to check `time` variable is set

Aster89:
Little change to make the code more modern.


da-edra (approved):
Thanks a lot! <3

PR #399, opened 2021-05-11
Remove note about i3blocks dynamic properties requirement

captn3m0:
v1.5 was released almost 2 years ago: https://github.com/vivien/i3blocks/releases/tag/1.5 with dynamic properties.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -72,4 +72,6 @@ Want to contribute?
 Great!
 Check the link:https://github.com/vivien/i3blocks-contrib/blob/master/CONTRIBUTING.md[contribution guidelines] in order to get started.
 
+NOTE: i3blocks-contrib REQUIRES i3blocks v1.5 or higher, since it uses i3blocks dynamic properties.
+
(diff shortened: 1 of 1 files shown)


da-edra:
Thanks for your contribution but unfortunately I'll have to close this PR, the reason for this is because users from Debian 🦖 and other very outdated Linux distributions only have i3blocks v1.4 in their repositories.

[Here](https://manpages.debian.org/buster/i3blocks/i3blocks.1.en.html) you can see the i3blocks version available in Debian's repositories, v1.5 is not even available in their `testing` and `unstable` repositories. 😞

captn3m0:
@da-edra How about a wording like:

>NOTE: i3blocks-contrib REQUIRES i3blocks v1.5 or higher, since it uses i3blocks dynamic properties.

da-edra:
@captn3m0 That sounds like an amazing idea! :)

da-edra (commented):
👍🏽

PR #402, opened 2021-05-18
iface: make display of wifi name optional and off by default

uykusuz:


iface/i3blocks.conf
```diff
@@ -5,3 +5,5 @@ command=$SCRIPT_DIR/iface
 #ADDRESS_FAMILY=inet6?
 color=#00FF00
 interval=10
+# set this to 1 to display the name of the connected WIFI interface instead of the IP address.
+display_wifi_name=0
```
uykusuz, line 9: a bit atypical perhaps. I made this variable all lower-case. Throughout the blocks upper-case and lower-case is mixed, that's a bit confusing. blocks-settings are lower-case so I thought I make this lower-case as well.

atreyasha:
@uykusuz thanks for this PR, this works for me :)

Also :+1: for setting `display_wifi_name=0` from the start, as it maintains the previous behaviour by default.

uykusuz:
@atreyasha nice, thanks for checking it! previous behavior...yeah, of course in hobby projects I forget to think about maintaining backwards compatibility :D

da-edra (approved):
Awesome! Thanks a lot for implementing backwards compatibility @uykusuz !

PR #404, opened 2021-05-22
eyedropper: add support for Sway, and for copying to the clipboard

majorde1:
Please let me know if you think it would be better to create a separate blocklet for the Sway version instead of configuring this one as an option.


majorde1:
Sorry, I intended to create a separate pull request, but accidentally added this second commit to my earlier pull request. I'm not sure how to fix it, but I will try to figure it out.

majorde1:
Ok, I got this pull request back to just the one intended commit again. Sorry about that.

da-edra (approved):
I really like this! Having it as a single blocklet makes a lot of sense!

Thank you!

PR #409, opened 2021-06-27
Use /tmp instead of /dev/shm to store path.

thiagowfx:
Why? It seems more reliable to use /tmp, as /dev/shm may not always be
present[1].

[1]: https://superuser.com/a/45509


da-edra (approved):
This makes a lot of sense! Thanks for submitting your PR!! 🥇

PR #410, opened 2021-07-01
Adding a weather block

runverzagt:
Data is pulled from NOAA so only works within the USA. 
Script uses an agency-specific naming convention which would allow for other agencies to be added in the future.


da-edra (approved):
Awesome!! Thank you so much!! 😸

PR #412, opened 2021-08-01
battery-poly: fix division-by-zero exception

basploeger:
I noticed div-by-zero exceptions in my ~/.xsession-errors.


da-edra (approved):
Nice catch!

Thanks a lot for the fix @basploeger !! 🥇

PR #416, opened 2021-08-25
Adds the block 'volume-pipewire'.

Zeioth:
This new block aims to replace 'volume-pulseaudio' in the future, but it needs to be tested for a while first.

## Paraphrasing README.md:

This is a fork of volume-pulseaudio, that supports pipewire, pulseaudio, and alsa. This is achieved by using the tool pactl instead of pacmd. So hopefuly this should be the future of this block.


da-edra (approved):
Awesome!! Thank you so much!! 😸

PR #423, opened 2021-09-18
volume-pipewire depends on alsa-utils

initialxy:
volume-pipewire uses amixer, which comes in also-utils package. Without it, it's not functional.


da-edra (approved):
Thank you!

PR #427, opened 2021-10-16
i3-focusdwindow cause high CPU usage

kaiwk:


denisse-dev (changes requested):
Hello, thanks for your pull request.

Unfortunately I can't merge it yet because the sleep value is hardcoded and users can't enable or disable it at will and some users may not want to have this enabled by default.

Could you please update your PR so users can enable it from their i3blocks config file? This feature should be disabled by default.

kaiwk:
Hey @da-edra, I have updated the PR, but I think it would be better to give `IDLE_SECONDS` a default value since we actually have a busy wait here for polling window title. Anyway, i documented that in `README.md`.

zDEFz:
proposing change in the logic this works in the first place, also to display more usable things like id, title or class. code comes later.

zDEFz:
managed to alleviate the high cpu issue and extended on it.
If you want add a onclick event to copy it to the clipboard.
Can now get either title or classname.

No sleep needed whatsoever.
You can put any interval value from i3blocks.

### i3-focused-window
```
#!/bin/bash
OPTIONS="Options:
    -class       --classname                Get classname of current window
    -title       --titlename                Get titlename of current window
    -win         --windowname              Get window name of current window
    -dim         --dimensions              Get window dimensions of current window
"

function print_usage() {
  printf "Usage:\n"
  printf "\n$(basename "$0") %s\n\n" "$OPTIONS"
}


function getClassname()
{
    classname="$(xdotool getactivewindow getwindowclassname)"
    echo "class=[$classname]"
}


function getTitlename()
{
    #get titlename
    titlename="$(xdotool getactivewindow getwindowname)"
    echo "title=[$titlename]"
}

function getWindowName()
{
    windowname="$(xdotool getactivewindow getwindowname)"
    echo "Win Name::$windowname"
}

function getWindowDimensions()
{
    windowdimensions="$(xdotool getactivewindow getwindowgeometry | grep Geometry | cut -d ':' -f2)"
    echo "Dim:$windowdimensions"
}

while true; do
  if [ -z "$1" ]; then
    print_usage
    exit 1
  fi
  case "$1" in
  -class | --classname)
    getClassname
    exit 1
    ;;
  -title | --titlename)
    getTitlename
    exit 0
    ;;
   -win | --windowname)
    getWindowName
    exit 0
    ;;
   -dim | --dimensions)
    getWindowDimensions
    exit 0
    ;;
  esac; done
```

### i3blocks
```
[i3-focused-window]
command=$SCRIPT_DIR/custom/i3-focused-window -class
interval=1
```

#469

PR #431, opened 2021-11-05
volume-pipewire: fix display with non-english system locale

dowe:
pactl honours the system locale (as opposed to pacmd), so the parsing
(e.g. for the ACTIVE state) returns an empty string. This leads to
always showing "Sound inactive".


REDVM:
This fixes a real problem and should be merged. volume-pipewire is not usable with OS configured in a language other than English.

PR #433, opened 2021-11-26
volume-pipewire: use default sink if inactive

psvenk:
If all of the sinks are inactive, fall back to the default sink instead of saying "audio inactive". This allows changing the volume while no audio is playing.


junaruga:
Agree. I want to check the volume level even when the audio is inactive.

PR #435, opened 2021-12-10
 volume-pipewire: add active PORT

kidpixo:
As in Title: waiting for [pactl: add format flag for JSON output (!497)](https://gitlab.freedesktop.org/pulseaudio/pulseaudio/-/merge_requests/497#note_0934b9201704b62cb1661b6eba196a814645e8ac) to be out, I just extended the existing volume-pipewire blocklet and add the Active Port variable `$PORT` to the format.


kidpixo:
@Zeioth @initialxy @dowe @psvenk just pinging all the volume-pipewire contributors

PR #442, opened 2022-01-08
Fixed memory2 ram calc

Ryan-Maia:
## Abstract

The value of ram was wrong in display, so I researched ways to make it right, and created two functions: `getFreeRam()` and `getTotalRam` wich will get the correct values about the avaliable and usage of memory ram 

### getFreeRam()

Execute the command `grep MemAvailable /proc/meminfo` wich returns the amount of memory not used in the system and make some filtering to get only the number without the text

### getTotalRam()

Execute the command `grep MemTotal /proc/meminfo` wich returns the total memory size and make some filtering to get only the number without the text

### DIFF
--- a/memory2/freeram.c
+++ b/memory2/freeram.c
@@ -1,0 +1,68 @@
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <ctype.h>
+
+long getFreeRam()
+{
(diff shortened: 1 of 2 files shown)


zDEFz (approved):
Reported value seems correct now.
Well. If I only could merge :)

PR #451, opened 2022-04-22
rofi-wttr: Weather display with forecast popup

lomky:
A minimal weather line with forecast pop up on click. Using [rofi](https://github.com/davatorium/rofi) for the pop up and [wttr.in](https://wttr.in/) as the weather interface.


denisse-dev (approved):
Thank you very much!! 🥇

PR #462, opened 2022-09-05
purpleair: update deprecated URL and add API key support.

majorde1:
This PR fixes the purpleair blocklet. PurpleAir's old json API was turned down, so the blocklet no longer works as written at HEAD.

PurpleAir's new API requires an API Key and has a different output format. This PR updates the API URL, updates the response parsing, and updates the documentation with information on obtaining and using an API Key.


majorde1:
Hello, just checking in on this PR review.

PR #463, opened 2022-09-10
battery-duo

Feliks24:
-module for displaying simple battery percentage for 1 or 2 batteries -using acpi
-made to replace battery-poly because of lack of speed and a bug -written in bash
-options to display simply or with symbols


Feliks24:
changed the code

PR #469, opened 2022-10-09
feat: more performant focused x window title script

andis-sprinkis:
Almost instant focued X window title output with less CPU usage.

Meant as a performance improvement upon i3-focusedwindow, but due to GNU coreutils dependency and some opinionated output format differences, decided to put this in a separate util.


zDEFz:
Do you consider to extend to also have title, class or ID?
I proposed my change. No CPU usage indebted on my proposal.
It is even faster.

PR #471, opened 2022-11-12
volume-pulseaudio: use [[ <cond> ]] instead of [ <cond> ]

Aster89:
A few lines below, `[[ <cond> ]]` is used indeed, so I think using `[ <cond> ]` was a typo.


denisse-dev (approved):
Thanks! :)

PR #473, opened 2022-11-27
brightness: Manage the label format on the label property.

junaruga:
This commit changes how to manage the label format, managing the label on the label property in the `config` file rather than managing the label format "[symbol]: " in `main.c`
It's helpful when aligning the label format with other blocklets.

You can check the modified README [here](https://github.com/junaruga/i3blocks-contrib/tree/wip/brightness-label/brightness).


ayjayt:
Hey, are you still using this?
Did you notice the infinite loop 100% CPU bug?

junaruga:
No, I am not using the i3blocks right now. I also didn't notice the bug.

PR #484, opened 2023-03-19
screenshot: catpure the screen using click

shivankgarg98:
This i3block button provides an easy way to take screenshots (snipping tool) with a single click. The captured screenshot is automatically copied to the clipboard and can also be saved in a file by clicking on an interactive notification (sent with dunstify) if enabled with the desired configurations. As snipping tools are growing popular in other OS, this lighweight simple script may proved to be a great alternative.

It adds a button to take a screenshot with Left or Right click to capture a region or complete screen respectively, and provides the functionality to save in clipboard and/or disk.
This button is very helpful and increase productivity especially when taking quick notes, documentation, sharing information using social media, or collobarating with colleagues etc.


shivankgarg98:
There is a small problem in this when creating a process in background. We will not be able to create another instance of screenshot until previous dunstify notification exits.
To fix this we can use
`command=i3-msg -q exec -- /path/to/screenshot/screenshot` butI face another problem here. The script is not taking reading input like BLOCK_BUTTON, IS_SAVE, SAVE_DIR from config. As a workaround, we can hardcode them with desired value in the screenshot script and use above command in i3blocks/config.

I created an issue regarding this - https://github.com/vivien/i3blocks/issues/470.

shivankgarg98:
Thanks to @gportay 's workaround suggestion. 
I am pleased to inform you that my snipping tool (or screenshot) utility is functioning as intended and without any issues.
I have thorough tested it and made sure ithere isn't any bug.
I am excited to contribute to the project and would be grateful if you could review my pull request and merge it if you find it satisfactory.
Happy to incorporate any feedback :)

PR #486, opened 2023-03-28
Update battery-poly to handle "Not charging" battery state

freeqaz:
My laptop, a Lenovo p15s, seems to output this state when it is fully charged. It broke the script because it hit the "raise error" case, so I am adding this here for myself and hopefully others!


Dalachowsky:
Had the same issue on my thinkpad T570. I'd suggest replacing triple `or` with single `if raw_status in ["Unknown", "Full", "Not charging"]:`

PR #494, opened 2023-07-10
arch-update: fix exit code 1 bug when using yay

AGuyWhoLikesTetris:
yay also exits with exit code 1 when no aur packages need to be updated. the code should ignore this case and move on.

### DIFF
--- a/arch-update/arch-update
+++ b/arch-update/arch-update
@@ -107,7 +107,14 @@ def get_aur_yaourt_updates():
     return aur_updates
 
 def get_aur_yay_updates():
-    output = check_output(['yay', '-Qua']).decode('utf-8')
+    output = ''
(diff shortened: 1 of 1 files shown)


erizo-dev:
Hi I have the same issue
I think I have fixed it on my system by copy-pasting the yaourt logic
By the way I think the flag should be added to the readme as I think yay is nowdays more popular than yaourt.

```python
def get_aur_yay_updates():
    try:
        output = check_output(['yay', '-Qua']).decode('utf-8')
    except subprocess.CalledProcessError as exc:
        if not (exc.returncode == 1 and not exc.output):
            raise exc
        return []
    if not output:
        return []
```
[edit] yes actually the same change suggested by the PR.
Thanks

PR #496, opened 2023-08-06
mediaplayer: use playerctl's formatting support

hseg:
In addition, this avoids crashing in cases where there is no valid artist name


hseg:
Redundant with #500 and #502

PR #449, opened 2022-02-18
brightness: re-init file descriptors set. Fix high CPU consumption

Lezh1k:
There is an issue on newest kernels with this droplet. CPU consumption - 100% . 
The reason is that select function returns result immediately. This happens because set of file descriptors is changed after first call. 
Fixed by re-initialization of descriptors set. 
Tested on Linux lezh1k-laptop 5.15.24-1-lts and on newest linux kernel.

PR #512, opened 2023-12-29
brightness: Stop busy waiting on Linux

ivanasen:
This was happening on Linux after the first call to `select` in the while loop.
The `time_to_wait` was reset to have `tv_sec == 0` by the `select()` call and this
would effectively make us busy wait which would cause high CPU usage.

This is how `select()` works on Linux as per the [man page](https://man7.org/linux/man-pages/man2/select.2.html#:~:text=On%20Linux%2C%20select()%20modifies%20timeout%20to%20reflect%20the%20amount%20of%20time)


ivanasen:
There already was a fix in: https://github.com/vivien/i3blocks-contrib/pull/449

PR #513, opened 2024-01-11
Add another possible value

ghost:


ghost:
okay, i see there's a bunch of identical fixes already

PR #549, opened 2026-04-05
volume-pipewire: add flat-list sink+port cycling on click

brkay54:
## Summary

- Replace sink-only cycling with a flat-list approach that enumerates all `(sink, port)` pairs across every available sink
- Left/right click now cycles through every distinct output (e.g. Bluetooth → Speaker → Headphones) instead of being stuck when multiple ports exist on a single sink
- Active port label is appended to the display name when a sink has more than one port (e.g. `ALC257 Analog/Headphones`)

## Problem

When a system has a single sink with multiple ports (e.g. built-in audio with Speaker and Headphones), the previous `set_default_playback_device_next` function found only one sink and did nothing on click.

## Solution

Added two functions:
- `get_all_outputs` — enumerates all `sink|port` pairs via `pactl list sinks`
- `set_next_audio_output` — finds the current `(sink, port)` in the flat list, advances by ±1, then calls `pactl set-default-sink` and/or `pactl set-sink-port` as needed

## Test plan

- [x] Left-click cycles forward through all outputs
- [x] Right-click cycles backward
- [x] Audio streams follow to the new output
- [x] Block label updates to reflect active port when sink has multiple ports
- [x] Works with Bluetooth sink connected alongside internal audio

### DIFF
--- a/volume-pipewire/volume-pipewire
+++ b/volume-pipewire/volume-pipewire
@@ -106,10 +106,60 @@ function set_default_playback_device_next {
     move_sinks_to_new_default $default_sink
 }
 
+function get_all_outputs {
+    pactl list sinks | awk '
+        /^\s*Name:/ {
+            if (sink != "" && num_ports == 0) print sink "|"
+            sink = $2; num_ports = 0; in_ports = 0
+        }
+        /^\s*Ports:/       { in_ports = 1 }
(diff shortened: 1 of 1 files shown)


brkay54:
@vivien need just a quick look, i have tested on my device it works.

